PR #22, opened 2026-02-13
feat: navigation lifecycle export + sync functionality jon/intorg-402

JonathanMatthey:
- When updating navigation through strapi - will be saved as json ( source of truth )
- manual sync / import scripts to update strapi from that json when needed
- update astro headers to use that navigation

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "start": "astro dev",
     "build": "astro build",
+    "preview": "astro preview",
     "astro": "astro",
     "format": "prettier --write '**/*.{js,mjs,ts,tsx,json,md,mdx,astro}' && eslint --max-warnings=0 --fix .",
     "lint": "prettier --check '**/*.{js,mjs,ts,tsx,json,md,mdx,astro}' && eslint --max-warnings=0 ."


cms/scripts/sync-navigation.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  for (const config of configs) {
+    await updateNavigation({
+      baseUrl: STRAPI_URL,
+      token: STRAPI_TOKEN,
+      apiId: config.apiId,
+      configPath: config.configPath,
+      label: config.label
+    })
+  }
```
Infi-Knight: nit: can be abstracted to another function

src/components/pages/SummitHeader.astro
Infi-Knight: This component is relatively complex, can we break it down with co-located CSS and functionality

cms/src/api/foundation-navigation/content-types/foundation-navigation/lifecycles.ts
```diff
@@ -0,0 +1,8 @@
+import path from 'path'
+import { createNavigationLifecycle } from '../../../../utils/navigationLifecycle'
```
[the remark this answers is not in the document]
JonathanMatthey: yeah will be part of a separate ticket :) also using tsconfig `@/ ` notation

package.json
```diff
@@ -5,34 +5,32 @@
   "scripts": {
     "start": "astro dev",
     "build": "astro build",
-    "preview": "astro preview",
```
Infi-Knight: preview should stay, it helps view the build output

PR #26, opened 2026-02-13
Import all eng blogs into v5

Anca2022:
## Summary
[INTORG-350](https://linear.app/interledger/issue/INTORG-350/import-all-eng-blogs-into-v5)

This PR 
- updates copy of all dev blogs
- ensures all required images are present
- changes image paths inside articles from `/developers/img/blog/...` to `/img/blog/...`
- aligns the styling of the blog landing page (`[...page].astro`) and individual blog article page (`[...id].astro`) with the current design on `interledger.org/developers`
- fixes navigation (links and pagination) for dev blog posts / blog landing page
- removes redundant exports from blog pages (`export const prerender = true`) 

This PR does not address all existing issues regarding Dev Blog post. The remaining items will be handled in separate PRs:
- adding `Tags component` to blog landing page - issue[ INTORG-407](https://linear.app/interledger/issue/INTORG-407/blog-view-add-tags)


JoblersTune:
Should probably edit this "Engineering blog" "Developer's Portal" wording out for foundation blog 
image

JoblersTune:
Tags are a separate issue right?

Anca2022:
Yes, sorry, I forgot to mention - The page for individual foundation articles `[...id].astro` will be part of another PR
This PR only handles Engineering Blog

src/utils/stripPagination.ts
```diff
@@ -0,0 +1,4 @@
+export default function stripPagination(path: string) {
+  const parts = path.replace(/\/$/, '').split('/')
+  return isNaN(Number(parts.at(-1))) ? path : parts.slice(0, -1).join('/')
+}
```
JoblersTune:
```suggestion
export default function stripPagination(path: string): string {
  if (!path || path === '/') return '/'

  const normalized = path.replace(/\/+$/, '')

  const parts = normalized.split('/').filter(Boolean)

  const last = parts.at(-1)

  if (last && /^\d+$/.test(last) && Number(last) > 0) {
    parts.pop()
  }

  return '/' + parts.join('/')
}
```

If used incorrectly this can return null instead of a path. If it get's broken we should either return home or come up with a sensible fallback. Since this is a generic function name (i.e. not specific to blogs) we should ensure we catch those kind of edge cases.

JoblersTune (approved):

PR #42, opened 2026-02-18
refactor(imports): simplify deep relative imports with aliases jon/intorg-377

JonathanMatthey:
## Summary
Replaces deep relative import paths across the app and CMS with clearer alias-based imports, and fixes Strapi lifecycle import resolution in compiled `dist` output.

## Related Issue
Fixes INTORG-377

## Changes
- Adds/uses alias-based imports (`@/` and targeted CMS alias usage)
- Simplifies blog/summit page imports and CMS serializer/sync imports
- Fixes CMS lifecycle imports to resolve correctly at runtime (`../../../utils`)
- Includes associated lockfile updates

## Manual Test
1. Run `npm run lint` in repo root
2. Run `cd cms && npm run build`
3. Run `cd cms && npm run develop` and verify Strapi loads without lifecycle module errors


Anca2022 (dismissed):
not sure why `bun.lock` and `package-lock.json` are part of this PR, but the rest looks good

cms/src/api/blog-post/content-types/blog-post/lifecycles.ts
```diff
@@ -6,10 +6,7 @@
 
 import fs from 'fs'
 import path from 'path'
-import {
-  gitCommitAndPush,
-  resolveTargetRepoPath
-} from '../../../../utils/gitSync'
+import { gitCommitAndPush, resolveTargetRepoPath } from '../../../utils'
```
Infi-Knight, line 9: can we migrate these as well?

cms/package-lock.json
Infi-Knight, line 1: we should remove this lockfile, bun.lock exists

JonathanMatthey: ✅

cms/src/api/blog-post/content-types/blog-post/lifecycles.ts
JonathanMatthey: @/... aliases here are TS-only right now. Strapi runs compiled Node code from cms/dist, and Node can’t resolve @ without extra runtime alias config, so it breaks with Cannot find module. That’s why I kept lifecycle imports relative (or via local ./utils re-export) for runtime safety.

Infi-Knight: Hmm, I think Bun natively resolves `tsconfig.json` `paths` at runtime, so `@` aliases might work without any extra config. Not suggesting we do that now, but it's something to keep in mind as an alternative if we can be sure of Node API compatibility

Infi-Knight (approved):

Issue #89, opened 2026-03-02
RFC: Implementing MDX to Strapi Reverse Sync for files containing JSX Blocks


Infi-Knight:
# RFC: Implementing MDX to Strapi Reverse Sync for files containing JSX Blocks

## Problem statement

Current `buildPagePayload` in `cms/scripts/sync-mdx/mdxTransformer.ts` sends page body as one `blocks.paragraph`.
That breaks block structure when MDX includes JSX components.

### Current broken payload shape

```json
{
  "content": [
    {
      "__component": "blocks.paragraph",
      "content": "Intro text.\n\n<AmbassadorGrid slugs={[\"andria-barret\"]} />\n\n<Blockquote source=\"Jane\">Great quote.</Blockquote>"
    }
  ]
}
```

### Why this is a problem

- Strapi dynamic-zone UI cannot edit these as real blocks.
- Relations are not reconstructed (`slugs` never become Strapi relation IDs).
- Round-trip is lossy.

---

## Scope for this draft

For now, this document is intentionally scoped to:

- `<Ambassador />` -> `blocks.ambassador`
- `<AmbassadorGrid />` -> `blocks.ambassadors-grid`
- `<Blockquote />` -> `blocks.blockquote`

Any selected approach should be extensible.

---

### High-level flow

```mermaid
flowchart TD
  A["MDX file body"] --> B["remark + remark-mdx parse"]
  B --> C{"Top-level node"}
  C -->|"mdxJsxFlowElement"| D["Map component name to Strapi __component"]
  C -->|"markdown nodes"| E["Accumulate markdown"]
  D --> F["Extract props / children"]
  F --> G["Resolve relations (slug -> documentId)"]
  E --> H["Emit blocks.paragraph when boundary reached"]
  G --> I["Ordered Strapi block array"]
  H --> I
  I --> J["POST/PUT to Strapi"]
```

### Example input MDX

```mdx
Some intro paragraph.

<AmbassadorGrid
  heading="Meet our ambassadors"
  slugs={['andria-barret', 'caroline-sinders']}
/>

<Blockquote source="**Jane Doe**, Acme Corp">
  Interledger changed how we think about open payments.
</Blockquote>
```

### Expected block output

```json
[
  {
    "__component": "blocks.paragraph",
    "content": "Some intro paragraph."
  },
  {
    "__component": "blocks.ambassadors-grid",
    "heading": "Meet our ambassadors",
[...]

PR #91, opened 2026-03-02
Jm/intorg 444 mdx components import

JonathanMatthey:
## Summary
Adds support for importing foundation page MDX components into Strapi as structured dynamic-zone blocks instead of treating the body as a single markdown blob.

This change:
- parses `<Paragraph>`, `<AmbassadorGrid>`, `<Blockquote>`, and plain text into Strapi blocks during sync
- resolves ambassador slugs to Strapi document IDs for ambassador grid relationships
- adds a reusable `Paragraph` Astro component and registers it in the page renderer so synced content renders correctly on the site

## Related Issue
Fixes INTORG-444

## Manual Test
Reviewer steps:
1. Run the MDX sync against a foundation page that includes `Paragraph`, `AmbassadorGrid`, and `Blockquote` components, such as `src/content/foundation-pages/about-us.mdx`.
2. In Strapi, confirm the imported `foundation-page` entry stores those sections as dynamic-zone blocks instead of one raw markdown block.
3. Verify ambassador grid blocks connect the expected ambassador entries by slug.
4. Start the site and open the synced page route to confirm paragraph content, blockquotes, and ambassador grids render correctly.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] PR title follows Conventional Commits (e.g. `feat(cms): sync foundation page MDX components as Strapi blocks`)
- [x] Linked issue included

### DIFF
--- a/cms/scripts/sync-mdx/mdxComponentParser.ts
+++ b/cms/scripts/sync-mdx/mdxComponentParser.ts
@@ -1,0 +1,95 @@
+/**
+ * Parses MDX body content into structured blocks for Strapi sync.
+ * Supports: <Paragraph>, <AmbassadorGrid>, <Blockquote>, and plain text.
+ */
+
+export type ParsedBlock =
+  | { type: 'paragraph'; content: string }
+  | { type: 'ambassadors-grid'; heading?: string; slugs: string[] }
+  | { type: 'blockquote'; quote: string; source?: string }
+
+/**
+ * Extract string attribute from JSX: heading="value" or heading='value'
+ */
+function extractStringAttr(
+  attrs: string,
+  name: string
+): string | undefined {
+  const regex = new RegExp(`${name}=["']([^"']*)["']`, 'i')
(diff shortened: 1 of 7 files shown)


cms/scripts/sync-mdx/mdxComponentParser.ts
Infi-Knight, line 1: Regex is the wrong abstraction here. MDX/JSX is not reliably parseable with regex once nesting, multiline attrs, expressions, or escaped characters are involved. We should switch to AST parsing (remark + remark-mdx) and operate on MDX AST nodes, not string patterns and as suggested by Stephan here https://github.com/interledger/interledger.org-v5/issues/89#issuecomment-3988945492 we support only agreed components and only static literals/arrays/objects. No dynamic expressions. Recommend we stop iterating on regex and implement the parser-based path now with strict scope and fail-on-error semantics.

Infi-Knight:
closing this in favor of AST approach

PR #93, opened 2026-03-03
feat: add MDX block parser foundation for JSX imports

Infi-Knight:
## Summary
Adds the foundational MDX block parser infrastructure for the JSX imports pipeline (MDX → Strapi dynamic-zone blocks). Follows the outline discussed in the [RFC](https://github.com/interledger/interledger.org-v5/issues/89)

- Introduces `mdxBlockParser.ts` : AST-based parser using remark + remark-mdx with a pluggable component handler registry
- Introduces `jsxExtract.ts `: typed JSX attribute extractors (getStringAttr, getBooleanAttr, getStringArrayAttr) with strict validation
- Introduces `parserErrors.ts` :  MdxParserError class with 7 error codes for hard-fail on unsupported syntax
- Introduces `types.blocks.ts` :  Strapi REST API payload types for all block components
- Adds remark, remark-mdx, mdast-util-mdx-jsx dependencies to cms/package.json
- No behavior change to existing sync pipeline; this is scaffolding only

## Related Issue
Fixes INTORG-454

## Manual Test

- added new tests
pnpm --dir cms test:sync-mdx

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [ ] Linked issue included

### DIFF
--- a/cms/scripts/sync-mdx/jsxExtract.ts
+++ b/cms/scripts/sync-mdx/jsxExtract.ts
@@ -1,0 +1,266 @@
+/**
+ * JSX attribute extraction utilities for MDX AST nodes.
+ *
+ * Provides safe, typed accessors for reading props from
+ * `mdast-util-mdx-jsx` JSX element nodes. All accessors
+ * throw `MdxParserError` on unexpected shapes so callers
+ * get strict, actionable failures.
+ */
+
+import type { MdxJsxFlowElement, MdxJsxAttribute } from 'mdast-util-mdx-jsx'
+import { MdxParserError, ParserErrorCode } from './parserErrors'
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/**
+ * Find a JSX attribute by name on an MDX JSX element.
+ * Returns `undefined` when the attribute is absent.
+ */
+function findAttr(
+  node: MdxJsxFlowElement,
+  name: string
+): MdxJsxAttribute | undefined {
+  return node.attributes.find(
+    (attr): attr is MdxJsxAttribute =>
+      attr.type === 'mdxJsxAttribute' && attr.name === name
+  )
+}
+
+// ---------------------------------------------------------------------------
+// Public API
+// ---------------------------------------------------------------------------
+
+/**
+ * Extract a string attribute value.
+ *
+ * @example
+ * ```ts
+ * // <Ambassador slug="caroline-sinders" />
+ * getStringAttr(node, 'slug')                        // → "caroline-sinders"
+ * getStringAttr(node, 'bio')                          // → undefined
+ * getStringAttr(node, 'slug', { required: true })     // → "caroline-sinders"
+ * getStringAttr(node, 'bio', { required: true })      // → throws MISSING_REQUIRED_PROP
+ * // <Ambassador slug={dynamicVar} />
+ * getStringAttr(node, 'slug')                         // → throws DYNAMIC_EXPRESSION
+ * ```
+ *
+ * @param node - JSX AST node
+ * @param name - Attribute name
+ * @param opts.required - When true, throws if the attribute is missing
+ * @returns The string value, or `undefined` if absent and not required
+ */
+export function getStringAttr(
+  node: MdxJsxFlowElement,
+  name: string,
+  opts: { required: true }
+): string
+export function getStringAttr(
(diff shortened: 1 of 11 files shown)


cms/scripts/sync-mdx/mdxBlockParser.ts
```diff
@@ -0,0 +1,199 @@
...
+    if (
+      node.type === 'mdxFlowExpression' ||
+      node.type === 'mdxTextExpression'
+    ) {
+      const expr = node as { value?: string; position?: Root['position'] }
+      throw new MdxParserError({
+        code: ParserErrorCode.DYNAMIC_EXPRESSION,
+        message: `Top-level expression "{${expr.value ?? '...'}}" is not supported.`,
+        line: node.position?.start.line,
+        column: node.position?.start.column
```
JonathanMatthey: this is repeated like 10 times... use a helper or make a fn

cms/scripts/sync-mdx/mdxBlockParser.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    // ESM import/export statements are not allowed — MDX content in this
+    // pipeline is converted to Strapi blocks, not executed as JS modules.
+    if (node.type === 'mdxjsEsm') {
+      const esm = node as { value?: string; position?: Root['position'] }
+      throw new MdxParserError({
+        code: ParserErrorCode.DYNAMIC_EXPRESSION,
+        message: `Import/export statements are not supported: "${esm.value ?? '...'}"`,
+        line: node.position?.start.line,
+        column: node.position?.start.column
```
JonathanMatthey: this is repeated like 10 times... use a helper or make a fn

cms/scripts/sync-mdx/mdxBlockParser.ts
```diff
@@ -0,0 +1,199 @@
...
+      message: `Failed to parse MDX: ${err instanceof Error ? err.message : String(err)}`
+    })
+  }
+
+  const blocks: ParsedBlock[] = []
+
+  // Walk top-level AST nodes
+  for (const node of tree.children) {
+    if (node.type === 'mdxJsxFlowElement') {
+      const jsxNode = node as MdxJsxFlowElement
```
JonathanMatthey: you can leave this in... but it's redundant, you checked if it was mdxJsxFlowElement, then you cast it

cms/scripts/sync-mdx/jsxExtract.ts
```diff
@@ -0,0 +1,266 @@
...
+}
+
+/**
+ * Extract a boolean attribute value.
+ *
+ * @example
+ * ```ts
+ * // <Ambassador showLinks />           → true  (valueless)
+ * // <Ambassador showLinks={true} />    → true
+ * // <Ambassador showLinks={false} />   → false
```
JonathanMatthey, line 127: but look these are just as much dynamicexpressions are variables or functions, and in mdx or even html they're represented as text then parsed.  so why not just skip the step of trying to parse, or throw dynamic expression errors and read everything as text ? much simpler

Infi-Knight: DYNAMIC_EXPRESSION for this parser means: any prop value that requires runtime evaluation.
So showLinks and showLinks={true|false} are static and valid; showLinks={someVar} / showLinks={fn()} are dynamic and must fail.

I don’t agree with silently degrading parse behavior. If a prop is invalid/dynamic, we should throw a parser error (with component/prop/line context), not coerce to undefined and let defaults hide bad MDX. That keeps import deterministic and preserves round-trip correctness.

If we want simplification, we can share extraction helpers, but strict per-type validation and hard-fail behavior should remain.

Infi-Knight: also MDX attrs are not “just text” semantically. In AST, literal, valueless, and expression are distinct shapes. So in my implementation the contract is strict/static parsing, not runtime evaluation. Dynamic expressions should fail fast. 
Treating everything as text introduces silent behavior drift:
- showLinks={someVar} used to hard-fail.
- Now it becomes undefined, field gets omitted, and Strapi default may apply (true) without warning.

Infi-Knight: keeping the strict behaviour as discussed on huddle

JonathanMatthey (approved):
Looks great 👏 🔥

PR #106, opened 2026-03-09
feat(cms): cleanup strapi admin UI INTORG-481

JonathanMatthey:
## Summary

- Disabled `draftAndPublish` on all Strapi content types — entries are always live when saved
- resstyled all the admin UI panels
- Added admin panel CSS to enforce a minimum textarea height of 140px - looks better 
- Hid the "Open Entity" item from the left sidebar navigation (content type still accessible via direct URL)

## Related Issue

Fixes INTORG-481

## Manual Test

1. Restart the CMS server
2. Open any content type — confirm no Draft/Published toggle appears
3. Open an Ambassador entry — i've restyled the layout..
4. Open a page editeo — restyled UI 
5. Confirm "Open Entity" is no longer visible in the left sidebar

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

screenshots of the admin UI changes

image
image
image


cms/src/admin/app.tsx
```diff
@@ -100,17 +100,48 @@ export default {
...
+      // Hide "Open Entity" from the left nav sidebar
+      document.querySelectorAll('div, span').forEach((el) => {
+        if (el.textContent?.trim() === 'Open Entity') {
+          const navItem = el.closest(
+            'a, li, [role="menuitem"]'
+          ) as HTMLElement | null
+          if (navItem) navItem.style.display = 'none'
+        }
+      })
+
```
Anca2022: I don't quite understand what is the purpose of `Open Entity`, but my question is: do we want to hide it for good, no one to be able to see it? or we just want to hide it from Editors, Authors? 
If the latter is true, we can edit access so that `Open Entity` is not available for editors / authors - they won't see it anymore

Image

cms/src/admin/app.tsx
```diff
@@ -100,17 +100,48 @@ export default {
...
+      const url = window.location.pathname
+      for (const [slug, title] of Object.entries(singleTypeTitles)) {
+        if (url.includes(slug)) {
+          const h1 = document.querySelector('h1')
+          if (h1 && /^[a-z0-9]{15,}$/.test(h1.textContent?.trim() ?? '')) {
+            h1.textContent = title
+          }
+          break
+        }
+      }
```
Anca2022: This does not seem to fix the issue, I still get the DocumentId as the page title for both `Summit navigation` and `Foundation Navigation`: 

Image

Anca2022: Plus, on live Stapi website the titles display correctly

Anca2022: There's an easier fix for this, directly from Strapi UI: 
go to `Content manager` - `Foundation Navigation` - left upper page corner - the three dots: `Configure view` - Entry title: change to `Id`

Image

cms/src/index.ts
```diff
@@ -161,116 +197,396 @@ async function configureFieldLabels(strapi: StrapiInstance) {
...
-      pageType: 'Page Type (Grant, Policy, Developer)',
+      path: 'Route Path',
+      pageType: 'Page Type',
       seo: 'SEO',
       hero: 'Hero',
-      content: 'Content',
-      createdAt: 'Created At',
-      updatedAt: 'Updated At',
-      publishedAt: 'Published At'
+      content: 'Page Content'
```
Anca2022, line 210: I still see the old labels - do they display correctly on your end? :-?

Image

cms/src/index.ts
```diff
@@ -161,116 +197,396 @@ async function configureFieldLabels(strapi: StrapiInstance) {
...
+    'navigation.menu-item': {
+      label: 'Label',
+      href: 'Link URL',
+      openInNewTab: 'Open in New Tab'
+    },
+    'shared.article-bio': {
+      author: 'Author Name',
+      profileBio: 'Author Bio',
+      profileImage: 'Profile Photo'
+    },
```
Anca2022, line 255: 🔥🔥🔥   
Image

cms/src/index.ts
```diff
@@ -161,116 +197,396 @@ async function configureFieldLabels(strapi: StrapiInstance) {
...
+    'blocks.image-row': {
+      heading: 'Heading',
       content: 'Content',
-      order: 'Display Order',
-      createdAt: 'Created At',
-      updatedAt: 'Updated At',
-      publishedAt: 'Published At'
+      image: 'Image',
+      imagePosition: 'Image Position',
+      attribution: 'Image Attribution'
```
Anca2022, line 348: This is great! :D 
Missing: `Ambassador`, `Blockquote` and `Callout Text`

cms/src/index.ts
```diff
@@ -161,116 +197,396 @@ async function configureFieldLabels(strapi: StrapiInstance) {
...
+      if (!meta) continue
+      const currentLabel = meta.edit?.label
+      const isDefault = !currentLabel || currentLabel === fieldName
+      if (isDefault && currentLabel !== label) {
+        updatedMetadatas[fieldName] = {
+          ...meta,
+          edit: { ...meta.edit, label },
+          list: { ...meta.list, label }
         }
+        needsUpdate = true
```
Anca2022: Here the logic is somehow off; it sets `needsUpdate` to be true even when no updates are needed
For example: simple save of `index.ts`, no changes made, triggers 
Image

Anca2022: Changing labels for Collections / Components does not update the labels in the UI for me right now

cms/src/admin/app.tsx
JoblersTune: We do have an issue to follow up on resolving this properly later too https://linear.app/interledger/issue/INTORG-403/remove-open-entity-collection-type-from-strapi
So let's not let this slow us down too much now

cms/src/index.ts
Anca2022: It might be something on my end, please let's talk when you get back :)

JonathanMatthey: works on my end 
image

cms/src/index.ts
JonathanMatthey: should be fixed - check again please :)

cms/src/index.ts
JonathanMatthey: done

cms/src/admin/app.tsx
JonathanMatthey: fixed

cms/src/admin/app.tsx
```diff
@@ -100,17 +100,58 @@ export default {
...
+      /* TEMP UI Fix: hide only the Preview aside (last one), not the one above */
+      aside[aria-labelledby="additional-information"]:last-of-type { display: none !important; }
+    `
+    document.head.appendChild(style)
+
+    // TEMP UI Fix: DOM polling for UI tweaks until Strapi exposes better extension points
     const interval = setInterval(() => {
-      // Find all buttons in the admin panel
+      // TEMP UI Fix: hide "Open Entity" from the left nav sidebar
+      document.querySelectorAll('div, span').forEach((el) => {
```
Infi-Knight: `querySelectorAll('div, span')` every 100ms is extremely broad. This matches potentially thousands of nodes on each tick just to find "Open Entity". Prefer a tighter selector (e.g. nav a, [role="menuitem"]) or, better, use a MutationObserver targeting the sidebar container.

JonathanMatthey: true :)

JonathanMatthey: fck me, and there's a forEach as well !! :D ouch

cms/src/admin/app.tsx
```diff
@@ -100,17 +100,58 @@ export default {
...
+    style.textContent = `
+      /* TEMP UI Fix: minimum textarea height */
+      textarea { min-height: 140px !important; }
+      /* TEMP UI Fix: hide only the Preview aside (last one), not the one above */
+      aside[aria-labelledby="additional-information"]:last-of-type { display: none !important; }
+    `
+    document.head.appendChild(style)
+
+    // TEMP UI Fix: DOM polling for UI tweaks until Strapi exposes better extension points
     const interval = setInterval(() => {
```
Infi-Knight: we are not clearing the interval here. A MutationObserver on document.body would be more efficient (fires only on actual DOM changes rather than polling blindly). If keeping the interval, it should at least clearInterval once all expected mutations have been applied. As written, it keeps querying and patching the DOM every 100ms for the entire admin session.

cms/src/admin/app.tsx
```diff
@@ -100,17 +100,58 @@ export default {
...
+      }
+      const url = window.location.pathname
+      for (const [slug, title] of Object.entries(singleTypeTitles)) {
+        if (url.includes(slug)) {
+          const h1 = document.querySelector('h1')
+          if (h1 && /^[a-z0-9]{15,}$/.test(h1.textContent?.trim() ?? '')) {
+            h1.textContent = title
+          }
+          // Fix browser tab title (document.title) when it shows document ID
+          const docIdPattern = /^[a-z0-9]{15,}\s*\|/
```
Infi-Knight: just to confirm, are strapi document IDs 15 or 24 char hex strings? If 24 tightening to /^[a-f0-9]{24}$/ avoids false positives on any legitimate heading that happens to be long and lowercase.

cms/src/admin/app.tsx
JonathanMatthey: done !

cms/src/admin/app.tsx
JonathanMatthey: extended to 20–26 character range check, alphanumeric check, lowercase/digit... should avoid false positives

Infi-Knight (approved):

PR #107, opened 2026-03-10
fix: align i18n routes and rename slug to pathSlug

JoblersTune:
## Summary

This PR standardizes content routing and naming by:

- switching localized summit/blog routes to locale-first URLs such as /es/summit/...
- renaming content/model references from slug to pathSlug
- updating sync, serializers, and language-switcher handling to match the new conventions

## Manual Test

All functionality should be identical. I did try testing out the import and export scripts successfully. 
Astro website still looks the same.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [ ] Linked issue included


JoblersTune:
Additional context

Please also pay special attention to what I may have missed on the cms end.

Strapi needs to use pathSlug in the frontmatter instead of slug.
Strapi needs to nest Spanish files appropriately based on our new file structure (this will likely effect both import and export scripts):
/foundation-pages/es
/summit-pages/es
/foundation-blog-posts/es
etc.

We'll need the same management for any collection, i.e. ambassadors as well.

Using `/summit-pages/es` instead of `/es/summit-pages` keeps the content types the same at the top-level grouping. That keeps all `summit-pages` entries together, makes collection-based tooling and queries more predictable, and lets locales act as variants of the same collection rather than separate top-level buckets.

Renamed slug to pathSlug because slug is reserved by Astro content collections and was causing filtering issues where Spanish entries overrode English entries when they shared the same value.

cms/scripts/sync-mdx/mdxTransformer.test.ts
```diff
@@ -478,15 +478,15 @@ describe('buildPagePayload', () => {
...
         locale: 'en',
-        resolveRelation: async (_apiId: string, slug: string) => ({
-          documentId: `doc-${slug}`
+        resolveRelation: async (_apiId: string, pathSlug: string) => ({
+          documentId: `doc-${pathSlug}`
         })
       }
 
       const mdx = createMdxFile({
         slug: 'ambassadors-page',
```
Anca2022:
```suggestion
        pathSlug: 'ambassadors-page',
```

Anca2022: You missed several `slug`s in this file: L487, L517, L535, L575, L610

cms/scripts/sync-mdx/mdxTransformer.ts
Anca2022, line 1: on line 96: "message: `[${mdx.slug}] ${err.message}`" -> `slug` should be renamed to `pathSlug`

cms/src/utils/flatContentLifecycle.ts
Anca2022, line 1: `cms/src/utils/blogLifecycle.ts` also needs replacing `slug` with `pathSlug`

src/utils/paths.ts
Anca2022, line 1: 'src/utils/static-paths.ts' needs updating

Image

cms/src/utils/flatContentLifecycle.ts
Anca2022: Also, for the moment BlogLifecycle doesn't handle localization, as it is not a `phase 1` issue => in case someone creates a Spanish blog post from Strapi admin, the lifecycle saves it in 'src/content/foundation-blog-posts' and doesn't add 'localizes' to frontmatter
The `sync:mdx` script only updates the english entries - for the same reason listed above
I think this should be a different issue, what do you think? And in the meanwhile maybe we should disable translating blog content from Strapi?

src/pages/es/blog/index.astro
```diff
@@ -4,15 +4,16 @@ import FoundationPageLayout from '@/layouts/FoundationPageLayout.astro'
 import { createExcerpt } from '@/utils/create-excerpt'
 
 // Only return posts with `draft: true` in the frontmatter
-const esEntries = await getCollection('foundation-blog', ({ data }) => {
+// TO DO get fopundation blog collection.
+const esEntries = await getCollection('developers-blog', ({ data }) => {
```
Anca2022: is this change correct? This should be the page for Foundation blog, right?

src/pages/es/blog/index.astro
```diff
@@ -81,7 +82,7 @@ const esEntries = await getCollection('foundation-blog', ({ data }) => {
                   {blogPostEntry.data.date.toDateString()}
                 </time>
                 <a
-                  href={`/blog/${blogPostEntry.id}`}
+                  href={`/es/blog/${blogPostEntry.id}`}
```
Anca2022:
```suggestion
                  href={`/blog/${blogPostEntry.id}`}
```
Otherwise we get 'es/blog/es/{title}'

Image

Anca2022: Also, it is not clear to me: we are only nesting Collections in this format: 
`{collection}/{filename}` and `{collection}/es/{filename}`

Or the aim is to also nest routing in the same way? 
`/blog/es/{article-name}`? 

Because if we also want to use this format for routing, then we need to restructure this `/pages/es` folder => we would rather want it to become `/pages/blog/es` and `/pages/developers/blog/es` instead, right?

src/pages/es/developers/blog/index.astro
```diff
@@ -81,7 +81,7 @@ const esEntries = await getCollection('developers-blog', ({ data }) => {
                   {blogPostEntry.data.date.toDateString()}
                 </time>
                 <a
-                  href={`/blog/${blogPostEntry.id}`}
+                  href={`es/developers/blog/${blogPostEntry.id}`}
```
Anca2022: Image

Anca2022 (commented):
Out of the trillion `slug`s, you missed a couple 😆 ; I replaced them locally, if it’s easier, I can commit and push.

Also, the routing nesting isn’t entirely clear to me: I left a comment in the review about it.

Other than that, everything looks great! 😁

src/pages/es/blog/index.astro
JoblersTune: Routing should all be under /es/everythin- else. So we want src/pages/es/**
Thanks for catching these bugs

src/pages/es/blog/index.astro
JoblersTune: I was manually reverting translation work that I'd done already. I must have copied the wrong file back here. Thanks!

JoblersTune:
> I replaced them locally, if it’s easier, I can commit and push.

Yes please, if it's already there that would be great! I'll wait for you to push and then I'll make the fixes for what you caught

.github/copilot-instructions.md
```diff
@@ -151,7 +151,7 @@ pnpm run build    # Production build
 - Includes `NODE_OPTIONS="--max-old-space-size=4096"` to handle OOM errors during build
 - Automatically restarts the `strapi.service` after rebuild
 
-Content published in the CMS automatically generates MDX files in `src/content/foundation-pages/` via lifecycle hooks. MDX generation is handled by `cms/scripts/sync-mdx/index.ts`.
+Content published in the CMS automatically generates MDX files in collection directories such as `src/content/foundation-pages/`, with localizations nested under the type directory (for example `src/content/foundation-pages/es/`). MDX generation is handled by `cms/scripts/sync-mdx/index.ts`.
```
Anca2022, line 154: **MDX generation is handled by `cms/scripts/sync-mdx/index.ts`.** 
- Not part of this PR, I know, but this is not accurate; 

MDX generation is handled by lifecycles using functions in `/src/utils` files (e.g, `blogLifecycle.ts`, `pageLifecycle.ts` etc)

`cms/scripts/sync-mdx/index.ts` handles synching the content back to Strapi

src/components/blog/BlogIndexDevelopers.astro
```diff
@@ -41,23 +45,23 @@ const path = '/developers/blog'
...
                 </h2>
                 <a
                   class="mt-space-3xs rounded border-[1.5px] bg-white px-1 text-step--1 no-underline"
-                  href={`${path}/es`}
+                  href={`/es${path}`}
                 >
                   ES
                 </a>
               </div>
             ) : (
```
Anca2022, line 61: For BlogIndexFoundation I deleted this part right here, that displays the ES for Spanish blog posts, as we currently don't have any Spanish blog post on Foundation

I guess I'm just making a note that the two blog views, Developers and Foundation, have different functionality in this regard, right now
 (Also, I guess this functionality will get removed once the Blog view displays just the blog posts under a specific locale)

Image

pnpm-workspace.yaml
```diff
@@ -3,5 +3,6 @@ packages:
 
 onlyBuiltDependencies:
   - '@parcel/watcher'
+  - better-sqlite3
```
Anca2022, line 6: 🚀 🚀 🚀

Anca2022 (approved):
This is great! 🎉 
I only have that comment regarding `.github/copilot-instructions.md`, but that is not technically part of this PR, so it's not a blocker.

src/components/blog/BlogIndexDevelopers.astro
JoblersTune: All of this will actually get updated in my PRs that will follow which build out a standard LanguageSwitcher component. So we can touch back then.

PR #124, opened 2026-03-12
feat(cms): use route path to dictate page folder structure in lifecycle

JonathanMatthey:
## Summary

Replaces the dual `path` + `pathSlug` field model with a single **full path slug** field across
foundation pages and summit pages. The path slug now encodes the complete URL path (e.g.
`grant/grant-for-web`), and the MDX sync system maps it directly to a nested directory structure
on disk.

- **Schema**: removed the separate `path` field; `pathSlug` is now a plain `string` (not a uid)
  representing the full URL path without a leading slash
- **File layout**: path segments before the last `/` become directories; the last segment becomes
  the filename. Localized files insert the locale folder before the filename:
  - `about-us` → `foundation-pages/about-us.mdx`
  - `grant/grant-for-web` → `foundation-pages/grant/grant-for-web.mdx`
  - `grant/grant-for-web` (es) → `foundation-pages/grant/es/grant-for-web.mdx`
- **Scanner**: rewrote `scanMDXFiles()` to walk directories recursively so nested locale files
  (e.g. `grants/es/`) are no longer invisible to the sync pipeline
- **Lifecycle**: new `resolvePageFilepath()` handles both nested (new) and flat (legacy) path
  formats for backwards compatibility during migration
- **Validation**: new `pathSlugSchema()` normalises pathSlug by stripping leading/trailing slashes
  across all content types
- **Tests**: added `scan.test.ts` covering flat, nested, mixed-locale, and edge-case paths

## Related Issue

Fixes INTORG-485

## Manual Test

1. Create a foundation page in Strapi with a flat slug (e.g. `test-page`) — confirm MDX is
   written to `foundation-pages/test-page.mdx`
2. Create a foundation page with a nested slug (e.g. `grants/test-grant`) — confirm MDX is
   written to `foundation-pages/grants/test-grant.mdx`
3. Add a Spanish localization — confirm it lands at `foundation-pages/grants/es/test-grant.mdx`
4. Delete the page — confirm both English and localized MDX files are removed
5. Confirm existing flat-slug pages still resolve correctly (legacy compatibility)

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


cms/scripts/sync-mdx/scan.ts
```diff
@@ -30,10 +33,10 @@ function scanDirectory({
     return mdxFiles
   }
 
-  // Read all files in the directory
+  // Read all files recursively in the directory
   let files: string[]
   try {
-    files = fs.readdirSync(baseDir)
+    files = fs.readdirSync(baseDir, { recursive: true }) as string[]
```
Infi-Knight: Bug: locale files are double-processed, causing phantom English Strapi entries

Making `scanDirectory` recursive causes all `es/` locale files to appear in the base scan. Then `scanMDXFiles` runs its locale pass on `es/` separately. Every locale file ends up in `mdxFiles` **twice**, but with different `isLocalization` values:

I verified this locally:
```
➜  cms git:(jm/intorg-485-pagepaths-folders) ✗ pnpm run sync:mdx:dry-run --force


> interledger-cms@0.1.0 sync:mdx:dry-run /Users/ravisoni/ILF/interledger.org-v5/cms
> tsx scripts/sync-mdx/index.ts --dry-run --force

🚀 MDX → Strapi Sync
==================================================
[dotenv@17.3.1] injecting env (8) from ../.env -- tip: 🔐 prevent building .env in docker: https://dotenvx.com/prebuild
🔗 Connecting to: http://localhost:1337
🔍 DRY-RUN MODE - No changes will be made


📁 Syncing ambassadors...
   Found 0 MDX files (0 invalid skipped)
   🗑️  [DRY-RUN] Would delete: ambassador (en)

📁 Syncing foundation-pages...
   Found 15 MDX files (0 invalid skipped)
   🔄 [DRY-RUN] Would update: about-us (en)
      📌 Matched via localizes: about-us: sobre-nosotros (es)
      🌍 [DRY-RUN] Would update localization: sobre-nosotros (es)
   🔄 [DRY-RUN] Would update: home (en)        ← appears twice (double-processed)
   ...
   ✅ [DRY-RUN] Would create: educacion (en)        ← Spanish slug created as English entry
   ✅ [DRY-RUN] Would create: servicios-financieros (en)   ← Spanish slug created as English entry
   🔄 [DRY-RUN] Would update: home (en)        ← duplicate
   ✅ [DRY-RUN] Would create: sobre-nosotros (en)   ← Spanish slug created as English entry
```

Infi-Knight (dismissed):

cms/scripts/sync-mdx/scan.ts
JonathanMatthey: should be fixed in latest commit

cms/scripts/sync-mdx/scan.ts
Infi-Knight, line 1: Bug?: Spanish pages with non-root `path` are synced as English entries

The export pipeline and the scan pipeline disagree on where Spanish files live when a page has a non-root `path`.

**Export** (`resolvePageFilepath`) writes a Spanish grants page to:
src/content/foundation-pages/grants/es/grant-for-web-es.mdx

```md
---
pathSlug: 'grant-for-web-es'
title: 'ES Grant for Web'
path: '/grants/'
locale: 'es'
localizes: 'grant-for-web'
---
```


**Scan** looks for Spanish files in two places only:
1. Base recursive scan skips any file with `es` in its path parts 
2. Locale scan only walks `src/content/foundation-pages/es/`

So `grants/es/grant-for-web-es.mdx` is skipped by (1) and missed by (2). It falls through to `syncEnglishEntry`, which hardcodes locale `'en'` and creates a phantom English Strapi entry.

Reproduced by creating `src/content/foundation-pages/grants/es/grant-for-web-es.mdx` with `locale: 'es'` and `localizes: 'grant-for-web'`, then running `pnpm run sync:mdx:dry-run --force`:

✅ [DRY-RUN] Would create: grant-for-web-es (en)
^^^
should be a Spanish localization of grant-for-web

JoblersTune:
Can you please document in the root README how to use the path field in the frontmatter so that people adding pages directly as files can understand the use case too.

cms/scripts/sync-mdx/scan.ts
JonathanMatthey: ok, fixed

image

JoblersTune:
I think pathSlug should be the single source of truth for the public route.

With [...page].astro, Astro already supports nested paths through a single rest param, so values like grant/financial-services work naturally.

Introducing a separate path field adds complexity without giving us much benefit, because Astro doesn’t have a clean way to model “arbitrary nested segments + final slug” as separate dynamic params.

In practice, a single nested pathSlug maps more directly to how Astro routing actually works.

This might also give a cleaner experience on Strapi if the user simply enters the whole url and we can extract the path from it. For example, summit pages would not have "summit" in the path since that is already defined on a file level. 

Perhaps we could make it better by adding helper text to show from where the path begins, like for foundation pages we say https://interledger.org/ and for summit pages it starts with https://interledger.org/summit and for Spanish foundation pages we have https://interledger.org/es/ and Spanish summit pages have https://interledger.org/es/summit/ and people just fill in the last bit. Then that's exactly what we need as the pathSlug

cms/src/utils/pageLifecycle.ts
```diff
@@ -91,15 +93,52 @@ export interface PageLifecycleConfig {
...
 }
 
-function getOutputDir(config: PageLifecycleConfig, locale: string): string {
-  const projectRoot = getTargetRepoRoot()
-  const baseOutputDir = path.join(projectRoot, config.outputDir)
+/**
+ * Resolves the MDX filepath for a page from `pathSlug` (full URL path, no leading slash).
+ * Segments before the last `/` are directories; the last segment is the filename stem.
+ *
+ * Legacy: if `path` is set (old Strapi), uses {outputDir}/{path}/{locale?}/{pathSlug}.mdx.
```
JoblersTune:
```suggestion
```

I don't think we need to reference that since it's not how it works going forward

cms/src/utils/pageLifecycle.ts
```diff
@@ -91,15 +93,52 @@ export interface PageLifecycleConfig {
...
+ *
+ * English: grant/ambassadors → {outputDir}/grant/ambassadors.mdx
+ * Spanish: grant/ambassadors → {outputDir}/es/grant/ambassadors.mdx
+ * English: about-us         → {outputDir}/about-us.mdx
+ */
+export function resolvePageFilepath(
+  outputDir: string,
+  page: Pick<PageData, 'pathSlug' | 'path'>,
```
JoblersTune: Surely we can delete mentions of `path`? Do we need to keep supporting them?

cms/src/utils/pageLifecycle.ts
```diff
@@ -91,15 +93,52 @@ export interface PageLifecycleConfig {
...
+  const normalized =
+    page.pathSlug == null
+      ? ''
+      : String(page.pathSlug)
+          .replace(/^\/+|\/+$/g, '')
+          .trim()
 
-  if (locale === 'en') {
-    return baseOutputDir
+  const legacyPrefix = (page.path ?? '').replace(/^\/+|\/+$/g, '').trim()
```
JoblersTune: Surely we don't need legacy support. This website is not live yet

cms/src/utils/pageLifecycle.ts
```diff
@@ -43,7 +43,9 @@ interface PageData {
   id: number
   documentId: string
   title: string
-  pathSlug: string
+  /** May be null on afterDelete in some Strapi versions / payloads */
+  pathSlug: string | null
+  path?: string
```
JoblersTune:
```suggestion
```
Surely we don't need this anymore

cms/src/utils/pageLifecycle.ts
```diff
@@ -111,7 +150,8 @@ function generateMDX(
   const locale = page.locale || 'en'
   const isLocalized = locale !== 'en'
   const { localizes, ...restPreserved } = preservedFields
-  // Use englishSlug (current English pathSlug) if provided, otherwise fall back to preserved localizes
+  delete restPreserved.path
```
JoblersTune: We can surely get rid of this with legacy code support removal too

cms/scripts/sync-mdx/index.ts
```diff
@@ -86,7 +86,7 @@ async function main() {
   console.log('='.repeat(50))
   console.log(`   ✅ Created: ${results.created}`)
   console.log(`   🔄 Updated: ${results.updated}`)
-  console.log(`   🗑️ Deleted: ${results.deleted}`)
+  console.log(`   🗑️  Deleted: ${results.deleted}`)
```
JoblersTune:
```suggestion
 console.log(`   🗑️  Deleted: ${results.deleted}`)
```

cms/scripts/sync-mdx/index.ts
```diff
@@ -86,7 +86,7 @@ async function main() {
   console.log('='.repeat(50))
   console.log(`   ✅ Created: ${results.created}`)
   console.log(`   🔄 Updated: ${results.updated}`)
-  console.log(`   🗑️ Deleted: ${results.deleted}`)
+ console.log(`   🗑️  Deleted: ${results.deleted}`)
```
JoblersTune:
```suggestion
  console.log(`   🗑️ Deleted: ${results.deleted}`)
```

cms/src/utils/pageLifecycle.ts
```diff
@@ -266,11 +332,64 @@ export function createPageLifecycle(config: PageLifecycleConfig) {
...
+        oldSlug
+      )
+    },
     async afterUpdate(event: Event) {
       const { result } = event
       if (!result) return
       if (shouldSkipMdxExport()) return
+
       const label = uidToLogLabel(config.contentTypeUid)
+      const locale = result.locale ?? 'en'
```
JoblersTune: Please can we keep defaultLang as a const that we use instead of typing it out

cms/src/utils/pageLifecycle.ts
```diff
@@ -249,6 +277,44 @@ async function exportAllLocales(
...
+  return `${documentId}:${locale}`
+}
+
+/** Exported for unit tests — Strapi passes locale on localized document updates. */
+export function readLocaleFromUpdateEvent(event: {
+  params?: {
+    locale?: string
+    documentId?: string
+    data?: { documentId?: string; locale?: string }
+    where?: Record<string, unknown>
```
JoblersTune: It is not clear to me what this parameter is for? Can we give it a better name?

cms/src/utils/pageLifecycle.ts
```diff
@@ -249,6 +277,44 @@ async function exportAllLocales(
...
+const pendingPathSlugChanges = new Map<string, string>()
+
+function pendingPathSlugKey(documentId: string, locale: string): string {
+  return `${documentId}:${locale}`
+}
+
+/** Exported for unit tests — Strapi passes locale on localized document updates. */
+export function readLocaleFromUpdateEvent(event: {
+  params?: {
+    locale?: string
```
JoblersTune, line 307: What is the difference between locale and data.locale, why do we need both?

cms/src/utils/pageLifecycle.ts
JoblersTune: Why do before update and after update determine locale differently?

cms/src/utils/pageLifecycle.ts
```diff
@@ -249,6 +277,44 @@ async function exportAllLocales(
   return filepaths
 }
 
+/** Cache old pathSlug before update so we can delete old files if it changes (per locale). */
+const pendingPathSlugChanges = new Map<string, string>()
```
JoblersTune: Do we need a module-level Map for this?

cms/src/utils/pageLifecycle.ts
```diff
@@ -266,11 +332,64 @@ export function createPageLifecycle(config: PageLifecycleConfig) {
...
+    }) {
+      if (shouldSkipMdxExport()) return
+      // Strapi v5: documentId is in params.data.documentId
+      const documentId =
+        event.params?.documentId ?? event.params?.data?.documentId
+      if (!documentId) return
+
+      const locale = readLocaleFromUpdateEvent(event)
+
+      // Per-locale pathSlug: only the updated locale's file moves. Using English
```
JoblersTune: This context doesn't feel relevant. It explains a problem that we're solving. We shouldn't need to keep an explanation of a fix.

cms/src/utils/pageLifecycle.ts
```diff
@@ -266,11 +332,64 @@ export function createPageLifecycle(config: PageLifecycleConfig) {
...
+      // here caused Spanish-only slug edits to stash the EN slug, so we deleted
+      // the wrong paths and left stale es/.../page.mdx files on every rename.
+      const existing = await fetchPublished(config, documentId, locale)
+      if (!existing?.pathSlug) return
```
JoblersTune, line 405: We should at least have a log if this happens, this would be unexpected behaviour, right?

JoblersTune: Or do we hit this update logic for pages that haven't been saved before? 
If this is expected behaviour maybe just add a comment to make it clear.

cms/src/utils/pageLifecycle.ts
```diff
@@ -266,11 +332,64 @@ export function createPageLifecycle(config: PageLifecycleConfig) {
...
+
+      const outputDir = getOutputDir(config)
+      const currentFilepath = resolvePageFilepath(outputDir, existing, locale)
+      let oldSlug = existing.pathSlug
+      if (fs.existsSync(currentFilepath)) {
+        const content = fs.readFileSync(currentFilepath, 'utf-8')
+        const { data } = matter(content)
+        if (data.pathSlug && typeof data.pathSlug === 'string') {
```
JoblersTune: When would this be false? Surely pathSlug is a required field in frontmatter? So, either please add a comment if this is expected behaviour, or add logging or error handling if this breaks our expectations.

JoblersTune:
The updated implementation feels like it’s compensating for inconsistent or unclear Strapi behaviour rather than working from a well-understood lifecycle contract.

If Strapi’s response genuinely varies in shape, that should be handled in a more structured and intentional way rather than branching across multiple fallback paths.

I’d be more comfortable if we documented the expected lifecycle event shapes and built the logic around that, instead of handling several possible cases defensively.

cms/src/utils/pageLifecycle.ts
JonathanMatthey: done

cms/src/utils/pageLifecycle.ts
JonathanMatthey: Strapi is inconsistent about where it puts the locale in update event payloads — it depends on the Strapi version, plugin version, and call site. The three sources cover three different shapes that Strapi has been observed to emit:

- params.locale — the "standard" shape; Strapi sets this at the top level when doing a localized update via the Document Service API
- params.data.locale — some call sites (especially bulk/publish operations or plugin hooks) embed the locale inside the mutation payload body instead
- params.where.locale — older-style filter-based updates where the locale is part of the where clause used to identify which document variant to update

So: one concept (target locale), multiple Strapi payload layouts; the helper normalizes them for stable behavior.

cms/src/utils/pageLifecycle.ts
JonathanMatthey: The difference is about what's available at each lifecycle stage:

- beforeUpdate — event.result doesn't exist yet (the write hasn't happened). The only source of truth is the incoming request parameters (event.params), which Strapi populates inconsistently — hence the readLocaleFromUpdateEvent helper that checks three different places.
- afterUpdate — event.result is the saved document returned by Strapi after the write. At this point, the locale is simply a field on the returned record (result.locale), so you can read it directly. Much cleaner.

cms/src/utils/pageLifecycle.ts
JonathanMatthey: removed

cms/src/utils/pageLifecycle.ts
JonathanMatthey: added comment 
The guard being false means a corrupted or manually-edited file.

cms/src/utils/pageLifecycle.ts
JonathanMatthey: improved comment

cms/src/utils/pageLifecycle.ts
JonathanMatthey: removed, it's to deal with old path slugs when deleting

cms/src/utils/pageLifecycle.ts
JonathanMatthey: updated Record with a more specific var name

JonathanMatthey:
> The updated implementation feels like it’s compensating for inconsistent or unclear Strapi behaviour rather than working from a well-understood lifecycle contract.
> 
> If Strapi’s response genuinely varies in shape, that should be handled in a more structured and intentional way rather than branching across multiple fallback paths.
> 
> I’d be more comfortable if we documented the expected lifecycle event shapes and built the logic around that, instead of handling several possible cases defensively.

Strapi has quite a few diff API, events where the data is different, when renaming, moving things, and importing, i'm trying to support each.  If you want it a different way i can revert last few changes and see if there's a better way, will take longer.

cms/src/utils/pageLifecycle.ts
Infi-Knight: > Document Service API → params.locale, bulk/plugin → params.data.locale, update filter → params.where.locale

has this beeen documented/commented somewhere in the code? If not, let's document this

cms/src/utils/pageLifecycle.ts
```diff
@@ -249,6 +279,44 @@ async function exportAllLocales(
...
 
+/**
+ * Strapi document service: `event.params.where` is the filter that selects
+ * which row to update (often includes `locale` for i18n). Same “where” idea as
+ * a query clause; the key name is Strapi’s, not ours.
+ */
+export type StrapiDocumentServiceUpdateWhere = Record<string, unknown>
+
+/** Exported for unit tests — Strapi passes locale on localized document updates. */
+export function readLocaleFromUpdateEvent(event: {
```
Infi-Knight, line 305: Jon's explanation from the review thread (Document Service API → `params.locale`, bulk/plugin → `params.data.locale`, update filter → `params.where.locale`) is exactly what a future reader needs — could we move that into a comment block above this function?

The fallback logic itself is fine. The ask is just to make it self-documenting rather than relying on the GH thread for context.

A `console.debug` for which path resolved the locale would also make production debugging much easier without changing behavior.

cms/src/utils/pageLifecycle.ts
```diff
@@ -266,11 +334,67 @@ export function createPageLifecycle(config: PageLifecycleConfig) {
...
+      // locale has a published version — export writes MDX from published data, so
+      // a pathSlug change must remove the old file. Draft-only / never-published /
+      // missing slug: no MDX path to reconcile, so skip.
+      const existing = await fetchPublished(config, documentId, locale)
+      if (!existing?.pathSlug) return
+
+      const outputDir = getOutputDir(config)
+      const currentFilepath = resolvePageFilepath(outputDir, existing, locale)
+      let oldSlug = existing.pathSlug
+      if (fs.existsSync(currentFilepath)) {
```
Infi-Knight: This fetches the published slug from Strapi (line 358), then reads the MDX file on disk to potentially override it. When would these two diverge?

If the answer is "manual edits or a failed export," a one-line comment documenting that would help. If there isn't a concrete, reproducible scenario, I'd lean toward trusting the Strapi API value and dropping the file read — it adds disk I/O on every save and is the main source of the complexity Sarah flagged in her comment.

cms/src/utils/pageLifecycle.ts
```diff
@@ -283,18 +407,27 @@ export function createPageLifecycle(config: PageLifecycleConfig) {
...
+          `[${label}] Skipping MDX delete: pathSlug missing on deleted document (documentId=${result.documentId})`
+        )
+        scheduleGitSync(label)
+        return
+      }
+
+      console.log(`🗑️  Deleting ${label} MDX for all locales: ${slug}`)
+
+      const outputDir = getOutputDir(config)
+      removeLocalizesFromLocaleFiles(slug, () => outputDir, label)
```
Infi-Knight: Nit: this callback ignores the `locale` parameter — `() => outputDir` always returns the base dir regardless of which locale `removeLocalizesFromLocaleFiles` is iterating over.

It happens to work because `localeMdxUtils.ts` now uses `readdirSync(dir, { recursive: true })`, so scanning from the base dir still finds all locale files. But it's fragile — if someone removes `{ recursive: true }` later, this breaks silently.

Safer:
```ts
removeLocalizesFromLocaleFiles(slug, (locale) => path.join(outputDir, locale), label)
```

cms/src/api/foundation-page/content-types/foundation-page/schema.json
```diff
@@ -22,15 +22,10 @@
       "pluginOptions": { "i18n": { "localized": true } }
     },
     "pathSlug": {
-      "type": "uid",
-      "targetField": "title",
-      "required": true,
-      "pluginOptions": { "i18n": { "localized": true } }
-    },
-    "path": {
       "type": "string",
```
Infi-Knight, line 25: `pathSlug` changed from `uid` (which enforces uniqueness in Strapi) to `string`. Two pages could now share the same `pathSlug` and silently overwrite each other's MDX files.

Worth adding `"unique": true` to the schema, or a `beforeCreate`/`beforeUpdate` validation check.

(Same applies to `summit-page/schema.json`.)

cms/scripts/sync-mdx/scan.ts
```diff
@@ -4,51 +4,73 @@ import matter from 'gray-matter'
 import type { ContentTypes } from './config'
 import type { MDXFile } from './mdxTypes'
 
-interface ScanOptions {
-  baseDir: string
-  locale: string
-  isLocalization: boolean
+// Non-English locale dir names. Mirrors LOCALES (excluding 'en') in src/utils/mdx.ts.
+// Update this set when new locales are added to the project.
+const LOCALE_DIR_NAMES = new Set(['es'])
```
Infi-Knight: This duplicates the locale list from `LOCALES` in `mdx.ts`. When a new locale is added, both must be updated manually.

Consider deriving it:
```ts
import { LOCALES, defaultLang } from '@/utils/mdx'
const LOCALE_DIR_NAMES = new Set(LOCALES.filter(l => l !== defaultLang))
```

cms/scripts/sync-delete-all.ts
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env node
```
Infi-Knight, line 1: **No confirmation prompt** — running without `--dry-run` immediately deletes all entries. A dangerous script like this should require an explicit `--confirm` flag or interactive prompt for destructive mode.

Infi-Knight (approved):

cms/src/utils/pageLifecycle.ts
JonathanMatthey: added

cms/src/utils/pageLifecycle.ts
JonathanMatthey: added 
image

cms/src/utils/pageLifecycle.ts
JonathanMatthey: you're right - simplified ! 

```      // Use Strapi’s published pathSlug only. Export always writes that value into
      // frontmatter and resolves the filepath from it, so disk should match; manual
      // MDX edits or a bad export are not special-cased here (avoids disk I/O every save).
      event.state.oldPathSlug = existing.pathSlug```

cms/src/api/foundation-page/content-types/foundation-page/schema.json
JonathanMatthey: done
image

PR #141, opened 2026-03-17
fix: alt text not syncing from MDX frontmatter to Strapi, and inline JSX images breaking in CKEditor.

Infi-Knight:
## Summary

Fix alt text not syncing from MDX frontmatter to Strapi, and inline JSX images breaking in CKEditor.

**Inline `` tags breaking in CKEditor**

When a blog post body contained `` tags with JSX attributes (e.g. `style={{ width: '50%' }}`), CKEditor5 escaped the entire tag as literal text breaking the image. `normalizeImages.ts` converts any `` to markdown `alt` before content reaches Strapi. JSX-only attributes (`style`, `className`, `loading`, etc.) are intentionally dropped since markdown has no equivalent.

**`featureImageAlt` / `thumbnailImageAlt` not syncing**

These fields were parsed from frontmatter by the Zod schema but silently dropped, never included in the Strapi payload. In Strapi, `alternativeText` lives on the upload file record, not the blog post, so the fix calls `PUT /upload/files/:id` after resolving the image URL to an ID.

**Ambassador `photoAlt` not syncing**

Same problem as above, applied to ambassadors. `buildAmbassadorPayload` now patches the photo upload file's `alternativeText` when `photoAlt` is present in frontmatter.

**Shared-image guard**

`alternativeText` is a single field on the upload file, shared across every post that references it. If two entries have different alt values for the same image, whichever syncs last wins, and the next export overwrites both frontmatter files with that value. The fix: `updateUploadAltOnce` skips and warns if a given upload ID has already been patched in the current sync run. A `Set<number>` per content type is created in `buildContentTypes` and closed over in each `buildPayload`, so it persists across all entries in a single run.

## Related Issue

Fixes INTORG-483

## Manual Test

1. Add `featureImageAlt: 'some alt'` to a blog post frontmatter. Run sync. Open that image in Strapi Media Library → confirm `Alternative text` updated.
2. Add `Caption` to a post body. Run sync. Open the post in Strapi → confirm content shows `Caption`, not escaped HTML.
3. Give two posts the same `featureImage` URL with different `featureImageAlt` values. Run sync. Confirm the console warns `⚠️ Skipping alt update` for the second post.
4. Add `photoAlt: 'Headshot'` to an ambassador MDX. Run sync. Confirm `Alternative text` updated on the photo in Strapi Media Library.

## Checks

- [ ] `pnpm run format`
- [ ] `pnpm run lint`

## PR Checklist

- [ ] PR title follows Conventional Commits (e.g. `feat: ...`, `fix: ...`)
- [ ] Linked issue included


cms/scripts/sync-mdx/mdxTransformer.ts
```diff
@@ -306,20 +355,39 @@ export async function buildBlogPayload(
...
+  if (thumbnailImage && parsed.thumbnailImageAlt) {
+    await updateUploadAltOnce(
+      strapiUploadContext.strapi,
+      thumbnailImage,
+      parsed.thumbnailImageAlt,
+      updatedAltIds,
+      mdx.pathSlug
+    )
+  }
+
```
Anca2022, line 381: I get this `405 error` when trying to run `pnpm run sync:mdx --force`
Image

cms/scripts/sync-mdx/mdxTransformer.ts
```diff
@@ -225,18 +226,45 @@ function nullOrValue(v: unknown): string | null {
...
+  if (updatedAltIds.has(id)) {
+    console.warn(
+      `   ⚠️  Skipping alt update for upload #${id} in "${pathSlug}" — already updated this run (shared image). Update alt text via Strapi Media Library instead.`
+    )
+    return
+  }
+  await strapi.updateUploadAlt(id, alt)
+  updatedAltIds.add(id)
+}
+
```
Anca2022, line 253: `updatedAltIds` is `Set(0) {}` on every run of the function

cms/scripts/sync-mdx/strapiClient.ts
```diff
@@ -199,6 +201,16 @@ export function createStrapiClient({
     })
   }
 
+  async function updateUploadAlt(
+    id: number,
+    alternativeText: string
+  ): Promise<void> {
+    await request(`upload/files/${id}`, {
+      method: 'PUT',
```
Anca2022: I don't think we can update uploads
Image

cms/scripts/sync-mdx/mdxTransformer.ts
Infi-Knight: sorry about that, I tested with post but switched to put as it's semantically correct verb, fixed

Anca2022 (approved):
Looks great! :D

PR #147, opened 2026-03-20
feat(cms): split image uploads into git-tracked originals and jon/intorg-497

JonathanMatthey:
## Summary

- **Recursive `sync:images`** under `public/uploads/img/original/`; nested paths are encoded as `__` in Strapi `name` (Strapi disallows `/` in multipart filenames).
- **Shared layout rules** (`imageLayoutPaths.ts` + `layout-paths.js`) for original vs optimized paths and subfolders.
- **Custom Strapi upload provider** (`strapi-upload-local-split`): main image `url` → `img/original/…`; derivatives → `img/optimized/…` (mirrored subdirs); **do not delete** `img/original/` when Strapi deletes an upload.
- **`uploadSplitImages`** copies masters into nested `original/…` using `originalMasterUploadsRelFromStorageName`.
- **`uploadFileStableUrlLifecycle`**: skip thumbnail/breakpoint rows; normalize `__` names / hash suffix.
- **CKEditor upload response**: `ckeditorUploadResponseSanitizer` + middleware in `cms/src/index.ts` strips `formats` when derivative URLs incorrectly sit under `img/original/` (fixes broken **srcset** / preview before reload).
- **ESLint**: override for `cms/providers/strapi-upload-local-split/**/*.js` (`strapi` global, allow `require`); removed unused code in provider and `uploadSplitImages.ts`.

## Related Issue

Fixes INTORG-497

## Manual Test

1. Run Strapi locally; upload an image in Media Library and confirm files land under `uploads/img/original/` (and optimized variants under `uploads/img/optimized/` as applicable).
2. In the rich text editor (CKEditor), insert/upload an image and confirm preview works immediately (no 404 on `srcset`).
3. Run `pnpm --filter cms run sync:images` (or your usual sync) against a tree with nested folders under `img/original/` and confirm Strapi receives expected entries.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Screenshots for UI changes (if applicable) — **N/A** (CMS/upload pipeline only; add screenshots if you document Media Library/CKEditor behavior)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ yarn-error.log*
 .netlify
 
 presentations/
+
+# Strapi-generated responsive images (regenerated locally from originals; see cms/docs/IMAGES.md)
+public/uploads/img/optimized/
--- a/cms/config/plugins.ts
+++ b/cms/config/plugins.ts
@@ -1,4 +1,10 @@
 export default () => ({
+  upload: {
+    config: {
+      // Images → optimized/ (gitignored). Admin uploads also copy masters to img/original/ (overwrite by slugged name). Use pnpm sync:images to bulk re-import from original/.
+      provider: '@interledger/strapi-upload-local-split'
+    }
+  },
   ckeditor: {
     enabled: true
   },
--- a/cms/package.json
+++ b/cms/package.json
@@ -17,6 +17,7 @@
     "sync:all:dry-run": "pnpm run sync:mdx:dry-run && pnpm run sync:navigation:dry-run",
     "sync:delete-all": "tsx scripts/sync-delete-all.ts",
     "sync:delete-all:dry-run": "tsx scripts/sync-delete-all.ts --dry-run",
+    "sync:images": "tsx scripts/sync-images.ts",
     "test:sync-mdx": "vitest run scripts/sync-mdx",
     "test:serializers": "vitest run src/serializers"
   },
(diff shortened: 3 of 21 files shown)


JonathanMatthey:
ready for review... :)

Anca2022 (commented):
There are a few things that are not clear to me, let's discuss this when you're back :)

Anca2022 (commented):
A few areas that might be worth improving:
- The `sync:images` script currently generates duplicates. Starting from an image in `original/foundation-blog`, it creates:
  - a renamed version in the same folder (e.g. `some-image.png` → `some_image.png`)
  - an additional copy in the `original/` folder
Then, each of these three images is linked to Strapi, resulting in three separate references: `foundation_blog__some-image`, `foundation_blog__some_image`, and `some_image`.

I think the script should instead generate a single normalized image only in the `original/` folder (e.g. `original/foundation-blog/some-image.png` → `original/some_image.png`), and only those images should be linked in Strapi.

- Another point to consider: when an image is deleted in Strapi, only the reference is removed, while the files remain in the Astro folders. This means that on the next sync, the images are uploaded again. At the moment, there doesn’t seem to be a way to detect that an editor intentionally deleted an image in Strapi.

- When adding an image inside a blog post, the generated image path is absolute (e.g. `localhost/.../image-url`). It might be better to use a relative path instead, to ensure it works consistently across different environments.

- The PR references `cms/docs/IMAGES.md`, but this file is not included in the PR.

JonathanMatthey:
TODO
- [x] when adding image through CKEditor, still getting localhost:... link but after saving post it updates correctly to relative link.  ideally should be local after inserting.
- [ ] review if deleting all original files is what we want... 
- [ ] running sync:images seems to re add images into strapi...
- [ ] make sure delete also deletes the optimized images

PR #156, opened 2026-03-24
feat(parser): implement paragraph merging in parseMdxToBlocks

Infi-Knight:
## Summary

Consecutive markdown nodes (paragraphs, headings, lists, tables) are now accumulated into a single `blocks.paragraph`, split only when a JSX component interrupts or EOF is reached. A blog post with 80 markdown nodes now produces 1 block instead of 80. without this, `serializeContent()` would wrap every node in a separate `<Paragraph>` component.

Also fixes a missing `@` path alias in `cms/vitest.config.ts` that was causing `mdxTransformer.test.ts` to fail.

## Related Issue

Fixes INTORG-514

## Manual Test

1. Run `pnpm --dir cms test:sync-mdx`
2. Verify merging: `## Hello\n\nSome text here.` produces 1 block, not 2
3. Verify splitting: markdown → `<Paragraph>JSX</Paragraph>` → markdown produces 3 blocks

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused 

### DIFF
--- a/cms/scripts/sync-mdx/mdxBlockParser.ts
+++ b/cms/scripts/sync-mdx/mdxBlockParser.ts
@@ -174,6 +174,20 @@ export async function parseMdxToBlocks(
 
   const blocks: ParsedBlock[] = []
 
+  // Buffer for consecutive markdown nodes. Flushed as a single
+  // blocks.paragraph when a JSX component interrupts or EOF is reached.
+  // This prevents a post with 80 markdown nodes from producing 80 blocks.
+  const pendingMarkdown: string[] = []
+
+  function flushPending(): void {
+    if (pendingMarkdown.length === 0) return
+    const merged = pendingMarkdown.join('\n\n')
+    if (merged.trim()) {
+      blocks.push({ __component: 'blocks.paragraph' as const, content: merged })
+    }
+    pendingMarkdown.length = 0
+  }
+
   // Walk top-level AST nodes
   for (const node of tree.children) {
     // Check for JSX — either a top-level flow element (tags on separate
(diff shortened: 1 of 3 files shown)


cms/scripts/sync-mdx/mdxBlockParser.ts
```diff
@@ -174,6 +174,20 @@ export async function parseMdxToBlocks(
 
   const blocks: ParsedBlock[] = []
 
+  // Buffer for consecutive markdown nodes. Flushed as a single
+  // blocks.paragraph when a JSX component interrupts or EOF is reached.
+  // This prevents a post with 80 markdown nodes from producing 80 blocks.
```
Anca2022, line 179: It currently works just for JSX component =  `<Paragraph>` 
For JSX components nested inside `<Paragraph>`s, they are imported as plain text in Strapi, not as different components

And because the code below works just fine in Astro, a developer might nest JSX component inside Paragraphs like this: 
Image

Image

 But then, in Strapi, they are imported like this: 

Image

Anca2022 (commented):
Discussion pending: We might want to import nested JSX components as Strapi components as well

Anca2022 (approved):
LGTM :D

cms/scripts/sync-mdx/mdxBlockParser.ts
Infi-Knight: will tackle this here https://linear.app/interledger/issue/INTORG-535/hard-fail-parser-on-nested-jsx-inside-paragraph

PR #163, opened 2026-03-26
feat: create script to import summit data and speaker images from Sessionize

Anca2022:
## Summary
Create script to import summit data and speaker images from Sessionize
- The script saves summit data as json files in `/src/data/sessionize/{year}-speakers.json` and `/src/data/sessionize/{year}-talks.json`
- imports speakers' profile images under `/public/img/sessionize-speakers/{year}`

## Related Issue
Use API calls to fetch summit data from Sessionize - [INTORG-468](https://linear.app/interledger/issue/INTORG-468/use-api-calls-to-fetch-summit-data-from-sessionize)

## Manual Test
Run script for 2025 (defaults to currentSummitYear)
```pnpm run sync:sessionize ``` 
or pass year argument to run the script for the other years
```pnpm run sync:sessionize -- 2022``` 

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### CODE REVIEW

File: scripts/sync-sessionize.ts [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+async function fetchAndSaveImage(url: string, name: string) {
+  try {
+    const response = await fetch(url)
+    if (!response.ok) {
+      throw new Error(
+        `Failed to fetch image: ${url}; Status: ${response.status} ${response.statusText}`
+      )
+    }
+
+    const ext = path.extname(url) || '.jpg'
```
Infi-Knight: is it possible that the url has query params like this: `https://sessionize.com/image/foo.jpg?w=400`. In that case this will fail, path.extname would return
   .jpg?w=400, not .jpg. This will produce broken filenames.

  Fix: Strip query params before extracting the extension:
  `const ext = path.extname(new URL(url).pathname) || '.jpg'`

Anca2022: You're right, thank you! :D 
I made the change

src/utils/sessionize.ts
```diff
@@ -0,0 +1,24 @@
+export const YEARS = ['2022', '2023', '2024', '2025'].sort()
```
Infi-Knight, line 1: YEARS is ['2022', '2023', '2024', '2025'] (`strings`), but sessionizeApiMap is typed as `Record<number, ...>`. Ideally, pick one type and be consistent: probably number for both, or string for both.

scripts/sync-sessionize.ts
```diff
@@ -0,0 +1,131 @@
...
+    )
+    process.exit(1)
+  }
+
+  const rawData = await fs.readFile(imgUrlFileSource, 'utf-8')
+  const speakersData: SessionizeSpeaker[] = JSON.parse(rawData)
+
+  await clearFolder(imgFilePath)
+
+  await Promise.all(
```
Infi-Knight: I think we should consider handling concurrency here. this fires all image fetches simultaneously. For years with
large number of speakers, this hammers Sessionize with parallel requests and could trigger rate limiting or connection errors. Consider batching (e.g., chunks of 5-10 concurrent requests).

scripts/sync-sessionize.ts
```diff
@@ -0,0 +1,131 @@
...
+    await fs.access(imgUrlFileSource)
+  } catch {
+    console.error(
+      `❌ Speaker JSON not found for YEAR ${YEAR}: ${imgUrlFileSource}`
+    )
+    process.exit(1)
+  }
+
+  const rawData = await fs.readFile(imgUrlFileSource, 'utf-8')
+  const speakersData: SessionizeSpeaker[] = JSON.parse(rawData)
```
Infi-Knight: should add error handling here for potential parse failure

src/utils/sessionize.ts
Anca2022: I'm gonna pick string, as I'm already using the values in YEARS some place else

Infi-Knight (approved):
functionally lgtm

PR #168, opened 2026-03-31
feat(blog): migrate all 92 blog posts to Paragraph-wrapped component …

Infi-Knight:
## Summary
Migrated blog posts to the component format

## Related Issue
Part of INTORG-524

## Manual Test

- run the sync script and see strapi editor has the Paragraph components

## Checks

- [ ] `pnpm run format`
- [ ] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/cms/scripts/migrate-blog-to-components.ts
+++ b/cms/scripts/migrate-blog-to-components.ts
@@ -1,0 +1,140 @@
+#!/usr/bin/env node
+
+/**
+ * Migrate blog post MDX files from raw markdown to <Paragraph>-wrapped components.
+ *
+ * For each blog post:
+ *   1. Read the MDX file, split frontmatter and body
+ *   2. Parse the body with parseMdxToBlocks() → structured blocks
+ *   3. Serialize blocks with serializeContent() → <Paragraph>-wrapped MDX
+ *   4. Write frontmatter + new body back to the file
+ *
+ * Usage (from repo root):
+ *   pnpm --dir cms tsx scripts/migrate-blog-to-components.ts [--dry-run]
+ *
+ * INTORG-524
+ */
+
+import fs from 'fs'
+import path from 'path'
+
+// Side-effect imports: register component handlers
+import './sync-mdx/paragraphHandler'
+import './sync-mdx/videoEmbedHandler'
+import './sync-mdx/blockquoteHandler'
+import './sync-mdx/calloutTextHandler'
+import './sync-mdx/pdfEmbedHandler'
+import './sync-mdx/ambassadorHandler'
+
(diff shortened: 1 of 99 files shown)


cms/scripts/migrate-blog-to-components.ts
```diff
@@ -0,0 +1,136 @@
...
+      const blocks = await parseMdxToBlocks(body, ctx)
+
+      if (blocks.length === 0) {
+        console.log(`   ⏭️  ${file} — no blocks produced, skipping`)
+        skipped++
+        continue
+      }
+
+      const serialized = serializeContent(
+        blocks as Array<{ __component: string; [key: string]: unknown }>
```
Anca2022: The cast `as Array<...>` doesn;t actually fix the type error; TS still complains. 
I suggest extending the `StrapiBlockBase` interface in `cms/scripts/sync-mdx/types.blocks.ts` by adding `[key:string]: unknown` 
That makes the cast unnecessary and handles the TS error

Anca2022 (commented):
On the first run, the `migrate-blog-to-components` correctly wraps the content in `<Paragraph>` blocks, but running it a second time normalizes the content (HTML entity decoding, bracket escaping). From what I can tell, these are semantic changes and don't affect rendering, but they will appear as changes in git

For a clean baseline, please run the script a second time and commit those changes too
Image

Anca2022:
Also, the PR  references fixing INTORG-524, but this PR doesn't fully resolve that ticket. 
INTORG-524 is a verification task meant to be completed once all the functionality is in place, so it should remain open after this PR merges. :)

Infi-Knight:
Addressed both review comments:

1. **`StrapiBlockBase` index signature**: Added `[key: string]: unknown` to `StrapiBlockBase` in `types.blocks.ts`, removed the `as Array<...>` cast from the migration script. `ParsedBlock[]` now naturally satisfies `serializeContent()`'s parameter type.

2. **Idempotent migration script**: Running the script a second time was producing diffs (HTML entity decoding `&amp;` → `&`, bracket escaping `[` → `\[`). Root cause: the paragraph handler was re-serializing children via `mdast-util-to-markdown`, which normalizes content. Fix: added `sourceText` to `ParserContext` so the paragraph handler can slice raw content from AST positions instead of re-serializing. Also fixed a missing blank line between frontmatter and body. Verified: running the script twice now produces zero diffs.

3. **PR description updated** : Changed "Fixes INTORG-524" to "Part of INTORG-524" so the ticket stays open after merge.

Anca2022 (approved):
The PR has conflicts that need to be resolved, but other than that, it looks good to me :D

PR #171, opened 2026-04-02
feat: handle translations for summit data

Anca2022:
## Summary

This PR adds translation support for dynamic Sessionize content (talks and speakers).

- **Populate `translationMap` with Sessionize data** — ensures the `LanguageSwitcher` correctly navigates between `en` and `es` pages for all summit talks and speakers
- **Fallback to En** — when no Spanish content is found, the pages fall back to English content 
- **update README** with info on Adding support for a new language

## Related Issue
Handle translations for summit data - Fixes [INTORG-475](https://linear.app/interledger/issue/INTORG-475/handle-translations-for-summit-data)

## Manual Test
1. Navigate to `/summit/{year}/talks` and use the **Language Switcher** to toggle between `en` and `es`. Repeat for `/summit/{year}/speakers`.
2. Click into individual talk pages (`/summit/{year}/talk/{talk-title}`) and speaker pages (`/summit/{year}/speaker/{speaker-name}`) and toggle languages on each.
3. On individual `es` pages, verify one of the following:
   - Spanish content is displayed, **or**
   - A disclaimer is shown if no translated content is available for that page.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/src/components/shared/TranslationDisclaimer.astro
+++ b/src/components/shared/TranslationDisclaimer.astro
@@ -1,0 +1,7 @@
+---
+
+---
+
+<p class="p-space-xs italic text-step-0">
+  Esta página aún no está disponible en español
+</p>
--- a/src/components/shared/TranslationDisclaimer.astro
+++ b/src/components/shared/TranslationDisclaimer.astro
@@ -1,7 +1,0 @@
----
-
----
-
-<p class="p-space-xs italic text-step-0">
-  Esta página aún no está disponible en español
-</p>
--- a/src/components/summit/SessionCard.astro
+++ b/src/components/summit/SessionCard.astro
@@ -1,9 +1,10 @@
 ---
-import type { SessionizeSupportedLocale, TalkPreview } from '@/types/summit'
+import type { TalkPreview } from '@/types/summit'
 import { generateSlug } from '@/utils/slug'
 import { truncateText } from '@/utils/text'
 import SessionSubheader from './SessionSubheader.astro'
-import { defaultLocale, type Locale } from '@/utils/i18'
+import { type Locale } from '@/utils/i18'
+import { getTranslation } from '@/utils/summit-talks-speakers'
 
 interface Props {
   talkPreview: TalkPreview
@@ -13,10 +14,7 @@ interface Props {
 
 const { talkPreview, basePath, lang } = Astro.props
 const href = `${basePath}/${generateSlug(talkPreview.title)}`
-const needsTranslation = lang !== defaultLocale
-const translation = needsTranslation
-  ? talkPreview[lang as SessionizeSupportedLocale]
-  : null
+const translation = getTranslation(talkPreview, lang)
 const description = translation?.description ?? talkPreview.description
 const title = translation?.title ?? talkPreview.title
 ---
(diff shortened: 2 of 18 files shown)


src/components/shared/TranslationDisclaimer.astro
```diff
@@ -0,0 +1,7 @@
+---
+
+---
+
+<p class="p-space-xs italic text-step-0">
```
Infi-Knight: We are hardcoding Spanish disclaimer. If another locale is added, this component would show a Spanish message on a non-Spanish page. Instead, accept a lang prop and select the message accordingly,

src/types/summit.ts
```diff
@@ -1,3 +1,5 @@
+export type SessionizeSupportedLocale = 'es'
```
Infi-Knight: this is a string literal union of one.  The cast `lang as SessionizeSupportedLocale` in multiple components will silently pass for any locale value at build time. If a third locale is added to Locale, accessing `entry[lang as SessionizeSupportedLocale]` with an unsupported locale would be `undefined`, which happens to fall through to the English fallback correctly. So it works, but the cast obscures the fact that only `es` is truly supported.

Consider a type guard or utility function like `getTranslation(entry, lang)` that encapsulates the cast and fallback. This would centralise the pattern that's currently repeated in SessionCard, SessionDetailPage ...

src/components/summit/SessionCard.astro
```diff
@@ -1,16 +1,24 @@
...
+  lang: Locale
 }
 
-const { talkPreview, basePath } = Astro.props
+const { talkPreview, basePath, lang } = Astro.props
 const href = `${basePath}/${generateSlug(talkPreview.title)}`
+const needsTranslation = lang !== defaultLocale
+const translation = needsTranslation
+  ? talkPreview[lang as SessionizeSupportedLocale]
+  : null
```
Infi-Knight: The same 3-line block appears in 4 components:
```
const needsTranslation = lang !== defaultLocale
const translation = needsTranslation ? entry[lang as SessionizeSupportedLocale] : null
const title = translation?.title ?? entry.title
```
consider abstracting this in a shared utility

src/components/summit/SpeakerDetailPage.astro
```diff
@@ -1,25 +1,33 @@
...
   year: string
-  sessionBaseHref: string
-  lang: Locale
 }
 
-const { entry, year, sessionBaseHref, lang } = Astro.props
+const { entry, year } = Astro.props
+const { routeLocale: lang } = Astro.locals
+const langPrefix = lang !== defaultLocale ? `/${lang}` : ''
+const sessionBaseHref = `${langPrefix}/summit/${year}/talk`
```
Infi-Knight: const sessionBaseHref = `${langPrefix}/summit/${year}/talk`

This hardcodes /summit instead of using `Astro.locals.currentBasePath`. Compare with SessionDetailPage which correctly uses currentBasePath.

Infi-Knight (approved):

src/components/shared/TranslationDisclaimer.astro
Anca2022: Removing `TranslationDisclaimer` for now, waiting for Sarah to merge the TranslationDisclaimer she works on, to use that instead

Anca2022: Tracked by https://linear.app/interledger/issue/INTORG-563/add-translationdisclaimer-to-summit-pages

PR #183, opened 2026-04-07
fix(cms): Astro HERO: background image + remove secondary CTA btns INTORG-572 INTORG-573

JonathanMatthey:
## Summary

- Adds a `backgroundImage` field to the Strapi `shared.hero` component, replacing the old secondary CTA buttons
- When set, renders the image as the hero section background; falls back to the existing gradient when absent
- Fixes a bug where deleting an image (or any hero/SEO field) in Strapi left the stale value in the MDX frontmatter

## Related Issue

Fixes INTORG-573

## Manual Test

1. In Strapi admin, open any Foundation or Summit page
2. Upload a background image in the Hero > Background Image field and publish — verify it renders on the page
3. Delete the image and publish — verify the gradient fallback renders and `heroImage` is gone from the MDX file

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused

### DIFF
--- a/src/pages/index.astro
+++ b/src/pages/index.astro
@@ -4,6 +4,7 @@ import FoundationPageLayout from '../layouts/FoundationPageLayout.astro'
 import Paragraph from '../components/blocks/Paragraph.astro'
 import { type Locale } from '@/utils/i18'
 import { HOME_SLUG } from '@/utils/routes'
+import { getHeroSectionStyle } from '@/utils/heroSectionStyle'
 
 const pageLang: Locale = 'en'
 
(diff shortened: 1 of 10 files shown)


JoblersTune:
Just to be clear our hero does use primary CTAs, so please don't remove primary CTA just secondary CTA

src/pages/index.astro
```diff
@@ -37,7 +40,13 @@ const pageDescription = metaDescription || undefined
 >
   <main class="pb-space-l">
     <section
-      class="bg-gradient-primary bg-no-repeat bg-cover bg-[position:bottom_right] min-h-[50vh] py-space-xl text-white flex items-center"
+      class:list={[
```
Infi-Knight, line 43: The exact same class:list + style pattern is now copy-pasted in four files:
FoundationContentPage.astro, SummitContentPage.astro, index.astro, and es/index.astro. 

**Suggestion**: Extract a <HeroSection> Astro component that encapsulates the conditional background logic. This would also make the gradient-vs-image decision a single source of truth.

Infi-Knight (approved):

src/pages/index.astro
JoblersTune: Styling right now should be considered temporary. Proper execution can wait until we get actual designs. Then we can keep things clean and avoid repetition etc.

PR #209, opened 2026-04-09
chore: update Ambassador and AmbassadorGrid Astro components

Anca2022:
## Summary
This PR : 
- Updates the structure of Astro components to reflect  Fellows on prod: https://interledger.org/grant/ambassadors
- Adds ` Fellowship Program` page 
- remove `showLinks` field from Strapi and prop in Astro components

Individual Ambassador ID pages will be addressed in a following PR

## Related Issue
Fixes INTORG-561

## Manual Test

Go to `grant/ambassadors` to see the AmbassadorsGrid 

image

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


src/utils/ambassadors.ts
```diff
@@ -0,0 +1,23 @@
...
+  type StrapiAmbassador
+} from '@/components/ambassadors/types'
+
+export const toAmbassadorData = (
+  entry: AmbassadorFrontmatterType | StrapiAmbassador
+): AmbassadorData => {
+  const photo = typeof entry.photo === 'string' ? entry.photo : entry.photo?.url
+  const photoAlt =
+    typeof entry.photo === 'string'
+      ? (entry as AmbassadorFrontmatterType).photoAlt
```
Infi-Knight: AmbassadorFrontmatterType.photo is string | null. When it's null, typeof null === 'object', so the logic falls into the else branch and reads entry.photo?.alternativeText which returns undefined, even if the frontmatter has a valid photoAlt set. 

photo being null is a normal, expected state for a frontmatter ambassador (photo just hasn't been uploaded yet). When it happens:

  typeof entry.photo === 'string'  // false — typeof null === 'object'
    ? (entry as AmbassadorFrontmatterType).photoAlt
    : entry.photo?.alternativeText // null?.alternativeText → undefined

  …so photoAlt from the frontmatter is silently dropped. Using a non-photo-shaped discriminator (e.g. 'tagline' in entry, or a kind parameter) would correctly route to the frontmatter's own photoAlt field regardless of whether photo is string or null.

Infi-Knight (approved):

src/utils/ambassadors.ts
Infi-Knight: Simplest fix: check 'photoAlt' in entry instead, that field only exists on the frontmatter type, so it's a reliable way to tell the two shapes apart.
```
const isFrontmatter = 'photoAlt' in entry
const photo = isFrontmatter ? entry.photo : entry.photo?.url
const photoAlt = isFrontmatter ? entry.photoAlt : entry.photo?.alternativeText
```

PR #227, opened 2026-04-14
Jm/intorg 618 images alt text

JonathanMatthey:
## Summary
Implemented alt text for all images

What changed:
- Added/updated lifecycle + sync logic so media `alternativeText` round-trips for:
  - feature images
  - thumbnail images
  - ambassador photos
  - hero/background images
  - article bio profile images
- Ensured nullable alt is supported so decorative images render `alt=""`.
- Updated Astro schema/rendering paths to read the synced alt fields from MDX frontmatter where applicable (`featureImageAlt`, `thumbnailImageAlt`, `photoAlt`, `heroImageAlt`, `articleBios[].imageAlt`).
- Removed redundant entry-level alt fields in Strapi content types and kept guidance to edit alt via media item edit modal.

## Related Issue
Fixes `INTORG-618`

## Manual Test
1. In Strapi, edit media alt text (pencil icon on selected image) for:
   - blog feature image
   - blog thumbnail
   - ambassador photo
   - hero background image
   - article bio image
2. Publish content and verify exported MDX frontmatter contains corresponding alt values.
3. Run sync/import path (MDX -> Strapi) with changed alt values and verify Strapi media `Alternative text` is updated.
4. Verify rendered site output contains expected `alt` values, and decorative/empty values render as `alt=""`.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


src/layouts/FoundationBlogLayout.astro
```diff
@@ -46,7 +46,7 @@ const longDate = dateObj.toDateString().substring(4)
             <div class="bg-[image:var(--feature-gradient)] w-full" />
             <img
               src={frontmatter.featureImage}
-              alt={frontmatter.featureImageAlt || frontmatter.title}
+              alt={frontmatter.featureImageAlt ?? ''}
```
Infi-Knight, line 49: I understand that the intent is to support decorative images (alt=""), but `??` only guards against `null/undefined`. It does NOT guard against empty string `''`. So this works for the decorative case. However, for any existing content where alt text hasn't been explicitly set in Strapi, all these images silently become decorative. 

Suggested fix: Keep the fallbacks but allow explicit empty string to override:


// Only treat as decorative if alt was explicitly set to empty
`alt={photoAlt !== undefined && photoAlt !== null ? photoAlt : name}`

Walk through the same three cases:

photoAlt = 'Headshot of Jane': not undefined, not null → use photoAlt → alt="Headshot of Jane"
photoAlt = '' : not undefined, not null → use photoAlt → alt="" (decorative, as intended)
photoAlt = undefined : fails the check → use name → alt="Jane Doe" (safe fallback)

Infi-Knight (approved):
left some suggestions

PR #230, opened 2026-04-14
fix(cms-sync): fail fast when strapi is offline

JonathanMatthey:
## Summary
- Added a shared Strapi health check for sync scripts.
- `sync:mdx`, `sync:navigation`, `sync:images`, and `sync:delete-all` now fail fast with a clear “start Strapi” message.
- `sync:all` now exits cleanly on child script failure without noisy stack traces.

## Related Issue
- Fixes INTORG-624

## Manual Test
1. Stop Strapi and run `pnpm --dir cms run sync:all --force` (expect clear error).
2. Start Strapi and rerun (expect normal execution).

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
Screenshot 2026-04-14 at 20 29 16

### DIFF
--- a/cms/scripts/ensureStrapiRunning.ts
+++ b/cms/scripts/ensureStrapiRunning.ts
@@ -13,22 +13,20 @@ export async function assertStrapiRunning(
   timeoutMs = 4000
 ): Promise<void> {
   const normalized = baseUrl.replace(/\/+$/, '')
-  const probeUrls = [`${normalized}/_health`, `${normalized}/admin`]
+  const probeUrl = `${normalized}/_health`
 
   let lastError: unknown = null
 
-  for (const url of probeUrls) {
-    try {
-      const response = await withTimeout(
-        (signal) => fetch(url, { method: 'GET', signal }),
-        timeoutMs
-      )
+  try {
+    const response = await withTimeout(
+      (signal) => fetch(probeUrl, { method: 'GET', signal }),
+      timeoutMs
+    )
 
-      // Any HTTP response means the server is reachable (even 401/403/404).
-      if (response.status >= 100) return
-    } catch (error) {
-      lastError = error
-    }
+    // Any HTTP response means the server is reachable (even 401/403/404).
+    if (response.status >= 100) return
+  } catch (error) {
+    lastError = error
   }
 
   const reason =
--- a/cms/scripts/ensureStrapiRunning.ts
+++ b/cms/scripts/ensureStrapiRunning.ts
@@ -39,4 +39,3 @@ export async function assertStrapiRunning(
       `Start Strapi first: cd cms && pnpm run develop`
   )
 }
-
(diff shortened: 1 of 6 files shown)


cms/scripts/ensureStrapiRunning.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  return promiseFactory(controller.signal).finally(() => clearTimeout(timeout))
+}
+
+export async function assertStrapiRunning(
+  baseUrl: string,
+  timeoutMs = 4000
+): Promise<void> {
+  const normalized = baseUrl.replace(/\/+$/, '')
+  const probeUrls = [`${normalized}/_health`, `${normalized}/admin`]
```
Infi-Knight: The /admin fallback is effectively unreachable.

const probeUrls = [`${normalized}/_health`, `${normalized}/admin`]

```js
  for (const url of probeUrls) {
    try {
      const response = await withTimeout(
        (signal) => fetch(url, { method: 'GET', signal }),
        timeoutMs
      )
      // Any HTTP response means the server is reachable (even 401/403/404).
      if (response.status >= 100) return
    } catch (error) {
      lastError = error
    }
  }
```

  The loop only moves on to /admin if the /_health iteration throws. If fetch resolves with any response, status >= 100 is
   true (HTTP status codes start at 100), so we return and never try /admin.

  **Scenario A**: Strapi is running

  GET http://localhost:1337/_health → Strapi responds. Depending on version it's usually a 204 (HEAD-style liveness) or a 405 Method Not Allowed for GET. Either way:

  - fetch resolves with a Response
  - response.status is 204 or 405, both >= 100
  - We return

  /admin is never probed. Fine, we got what we needed.

  **Scenario B**: Strapi is down (port closed)

  GET http://localhost:1337/_health → TCP connect refused.

  - fetch rejects with TypeError: fetch failed (cause: ECONNREFUSED)
  - We hit the catch, save lastError, continue to /admin
  - GET http://localhost:1337/admin → same host, same port, same ECONNREFUSED
  - fetch rejects again
  - Loop ends, we throw

  So the /admin probe did run, but it told us nothing new. Both URLs share the host:port pair; if one can't connect, neither can.

  **Scenario C:** DNS failure / wrong host

  Same as B. Both probes fail identically on ENOTFOUND for the shared hostname.

  When would /admin actually save us?

  You'd need a case where:
  1. GET /_health throws (network-layer failure), AND
  2. GET /admin succeeds against the same host:port

  That requires a path-specific transport failure e.g., a proxy or WAF that drops the /_health path at the TCP level (resets the connection) but proxies /admin normally. Extremely unusual, and if you actually had that setup, the status >= 100 check on /admin would pass (it's a 302 redirect to the admin UI in Strapi), so yes, the fallback would work but
   this is a contrived edge case, not a realistic one against a local pnpm run develop.

So In the two scenarios that matter here, Strapi up and Strapi down the fallback never changes the outcome:
  - Up: /_health returns, we exit before trying /admin.
  - Down: both fail the same way.

Either drop it to a single probe, or keep both with a comment explaining the reverse-proxy edge case you're hedging against.

Infi-Knight (approved):
might not need the /admin probe

cms/scripts/ensureStrapiRunning.ts
JonathanMatthey: good catch, dropped /admin

PR #245, opened 2026-04-17
fix(nav): active state and keyboard accessibility [INTORG-629]

JonathanMatthey:
## Summary

Fixes two nav issues on the Foundation header:

1. **Active state** — the nav previously highlighted every item whose path was a prefix of the current URL. Visiting `/developers/blog` lit up both "Technology" (contains `/developers`) and "Content Hub". Switched to a longest-prefix-match strategy so only the most specific item is marked active.

2. **Keyboard accessibility for mobile nav** — at narrow viewports, the nav wrapper was still in the tab order even when slid off-screen, so tabbing could land on invisible links. Pressing Space on the hamburger opened the menu but left focus on the button, with no way to Tab in. Fixed by:
   - Adding `inert` to the nav wrapper in its initial closed state (both Foundation and Summit headers)
   - Toggling `inert` in `setOffscreenState` so closed = unreachable, open = reachable
   - Moving focus to the first menu item when the menu is opened

## Related Issue

Fixes INTORG-629

## Manual Test

1. On a Foundation page at a narrow viewport, open the hamburger menu with Space — focus should move to the first menu item and Tab should cycle through all items
2. Visit `/developers/blog` — only "Content Hub" should be underlined, not "Technology"
3. At wide viewport, Tab through the header — all nav links should be reachable in order

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


src/scripts/header-nav.ts
```diff
@@ -51,13 +64,63 @@ export function initHeaderNav(navId: string, iconId: string) {
...
+  let bestMatchLength = 0
+
+  for (const link of navLinks) {
+    const linkPath = link.pathname.replace(/\/$/, '')
+    // Skip the root path — the home link shouldn't activate from every subpage.
+    if (linkPath === '' || linkPath === '/') continue
+    if (currentPath === linkPath || currentPath.startsWith(linkPath + '/')) {
+      if (linkPath.length > bestMatchLength) {
+        bestMatch = link
+        bestMatchLength = linkPath.length
```
Anca2022, line 123: `bestMatch` works well for most pages, but there are 2 exceptions - ID pages for Summit Talks and for Summit Speakers - these shouldn't be under `Overview`, but under the `Sessions` and `Speakers` nav items, respectively

I think we should change the ID pages route to use plural form (`speakers`/`talks` instead of `speaker`/`talk`). 
that would fix this issue and would keep routing consistent. What do you think? 

Image

Anca2022 (commented):
Clicking outside the hamburger menu still doesn't close it on either the foundation or summit pages :)

And since you are working on the hamburger menu, could you please add the `ctaLink`s (“Interledger Foundation” / “Interledger Summit”) into it? Right now, those links disappear entirely on small screens.

src/scripts/header-nav.ts
JoblersTune: Add to redirects on v5 for all EN and ES previous singular links

Anca2022 (approved):
lgtm :)

PR #263, opened 2026-04-29
feat(styles): set up Figma design system tokens [INTORG-643]

Infi-Knight:
## Summary

Sets up the new Figma design system tokens (spacing, padding, radius, opacity, colours, typography) alongside the existing fluid scale. Nothing existing is changed: old utilities (`p-space-l`, `text-step-0`, `rounded-card`, etc.) keep working. New design migration, pillar colour remap, button refactor, and Titillium retirement happen in follow-ups (or get dropped entirely if design wants something different).

Things worth flagging that aren't obvious from `git diff`:

- `--spacing-3xl` (48px) and `--spacing-3xl-tight` (40px) coexist on purpose. Figma's `padding-3xl` is 8px tighter than `spacing-3xl` and Radu confirmed it's intentional. The README explains when to pick each.
- Typography uses Tailwind v4's `--text-{name}--line-height` / `--font-weight` companion-key pattern, so one `text-h1` class sets all three properties together. Standalone `--leading-*` tokens are also defined for cases where you only want the line-height (e.g. `text-h3 leading-h2`).
- `md:` and `lg:` breakpoints moved from 800 / 1000 to 768 / 1024 to match Figma's Tablet/Desktop tiers and Tailwind v4 defaults. Existing `md:`/`lg:` variants now snap 32px earlier and 24px later respectively. 
- Poppins is loaded but `body { font-family }` stays Titillium until the migration. Opt in per element via `font-poppins`.

[src/styles/README.md](src/styles/README.md) covers token usage, the 3xl-tight rule, responsive typography, leading utilities, and the companion-key mechanism.

## Related Issue

[INTORG-643](https://linear.app/interledger/issue/INTORG-643/set-up-design-system)

## Manual Test

The visible site should look identical to `main`.

1. `pnpm install && pnpm dev`, walk side by side with `main`:
   - Foundation home + a content page
   - Blog index + a single article (prose, table, code)
   - Summit page (header, session card, speakers list)
   - Ambassador grid + detail
   - Lander layout
   - A docs route under `/developers/*` (separate CSS lane, should be untouched)
2. Spot-check the new utilities via DevTools or a temporary element:
   ```html
   <div class="font-poppins text-h1 md:text-h1-md lg:text-h1-lg bg-deep-teal-100 text-neutral-0 p-3xl-tight rounded-2xl">
     test
   </div>
   ```
[...]

PR #256, opened 2026-04-22
feat(analytics): auto-instrument inline links with umami events [INTORG-487]

JonathanMatthey:
## Summary

- **Strapi content links** — Auto-adds `data-umami-event` on every `<a>` rendered from Strapi, so editors don’t have to tag links manually. Event strings look like `{pageLabel} link - {linkText}` (e.g. `Ambassadors page link - Jane LinkedIn`).

- **Paragraph / markdown** — Removed `markdown-it` from `Paragraph.astro`; everything goes through `marked` now. `parseMarkdown` / `parseMarkdownInline` take an optional umami context. `Paragraph`, `CalloutText`, and `Blockquote` accept a `umamiContext` prop.

- **Section** — `Section.astro` instruments links in the JSON walker. Also escaped `href` / text (small injection hole) and removed a dead guard.

- **Performance** — Cached `Marked` instances by page label to avoid re-allocating per block.

- **Tests** — Added Vitest at the repo root with 16 tests.

## Related Issue

INTORG-487

## Manual Test

1. `pnpm start` and open a page that renders Strapi content (e.g. a Callout, a Section, or a Paragraph block).
2. Inspect an inline link — confirm `data-umami-event="{Page} page link - {Link text}"` is present.
3. In the Umami dashboard, click an inline link and confirm the event registers with the expected label.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] PR title follows Conventional Commits (e.g. `feat: ...`, `fix: ...`)
- [x] Linked issue included
- [x] Scope is focused (target ~10-20 files when possible)
- [x] Screenshots for UI changes (if applicable)

PR #266, opened 2026-05-04
feat(buttons): add Button and LinkButton with CVA variants [INTORG-705]

Infi-Knight:
## Summary

Ships `Button.astro` and `LinkButton.astro` plus a long-lived SEO-excluded preview route at `/preview/ui-components`. Establishes the pillar-aware primary-button colour system through three runtime CSS vars (`--color-button-primary`, `--color-button-primary-hover`, `--color-button-primary-disabled`) with orchid as the `:root` default.

## Related Issue

Fixes INTORG-705. Builds on INTORG-643 (PR #263) which landed the design-system token foundation.

## Manual Test

1. `pnpm dev` and open http://localhost:1103/preview/ui-components
2. Visual matrix: every section renders (Primary lg/sm, Secondary lg/sm in light + dark, LinkButton, Form, RTL). All buttons should have rounded corners (8px), Poppins font, orchid background on primary.
3. Hover: primary darkens to #662b81. Secondary-light goes black-on-white. Secondary-dark goes white-on-black.
4. Tab through the page: every focused button shows a 2px solid orchid focus ring with 2px offset.
5. LinkButton aria-disabled: not focusable via Tab, clicking does nothing (href is stripped).
6. Pillar override (Pattern A) via DevTools console:
   ```js
   document.documentElement.setAttribute('data-pillar', 'tech')
   ```
7. Inline-style override (Pattern B) via DevTools:
   ```js
   document.querySelector('button').setAttribute(
     'style',
     '--color-button-primary: red; --color-button-primary-hover: darkred;'
   )
   ```
   First button turns red.
8. Class override (Pattern C): adding `class="w-full"` to a `<Button>` produces a full-width button (tailwind-merge lets caller classes win).

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused 

## Why CVA

Multi-axis variant matrix (variant x mode x size x iconOnly) handled with `compoundVariants`. Every Tailwind class is a literal string at definition time so the v4 content scanner sees them. Caller `class` flows through tailwind-merge so per-property dedup picks the right winner.

## Pillar-aware primary, how it works

Primary button colour resolves through three runtime CSS vars defined on `:root` in `src/styles/base/variables.css` and exposed as Tailwind utilities (`bg-button-primary`, etc.) via an `@theme inline` block in `src/styles/theme.css`. Orchid is the `:root` default, effectively "the orchid pillar." Per-pillar mappings for tech, mission, vision, and values are out of scope for this PR (blocked on design); each future pillar is a 3-line addition under a `[data-pillar='X']` selector.

## Caller overrides, three patterns

**Pattern A** (preferred when re-tinting to a known pillar):
```astro
<div data-pillar="tech">
  <Button>Tech-tinted action</Button>
</div>
```

**Pattern B** (preferred when forcing a specific palette colour):
```astro
<Button
  style="
    --color-button-primary: var(--color-orchid-100);
    --color-button-primary-hover: var(--color-orchid-150);
    --color-button-primary-disabled: var(--color-orchid-50);
  "
>
  Force orchid regardless of pillar
</Button>
```

**Pattern C** (preferred for one-off non-colour tweaks):
```astro
<Button class="w-full">Full-width primary</Button>
```


src/styles/base/variables.css
```diff
@@ -17,6 +17,15 @@
        @theme inline reads this; [data-pillar] selectors override it below. */
     --color-primary: oklch(51.54% 0.088 194.77);
 
+    /* Pillar-aware button colour. Orchid is the :root default, effectively
+       "the orchid pillar." Future pillars override these three vars under
+       [data-pillar='X'] selectors (one block per pillar, mirroring the
+       --color-primary pattern below). Callers can also override per-button
+       via Pattern B in docs/plans/intorg-705.md §4.6 (inline style on the element). */
```
Anca2022: nit: `docs/plans/intorg-705.md` do not exist on this PR

src/components/shared/buttonVariants.ts
```diff
@@ -0,0 +1,84 @@
...
+    'inline-flex items-center justify-center',
+    'font-poppins text-caption',
+    'rounded-lg select-none',
+    'focus-visible:outline-solid focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-orchid-100',
+    'motion-safe:transition-colors motion-safe:duration-200',
+    'disabled:cursor-not-allowed aria-disabled:cursor-not-allowed aria-disabled:pointer-events-none'
+  ],
+  {
+    variants: {
+      variant: {
```
Anca2022, line 30: Bug: for the secondary variant, the disabled button changes on hover

src/pages/preview/ui-components.astro
```diff
@@ -0,0 +1,410 @@
...
+            fill="none"
+            aria-hidden="true"
+          >
+            <path
+              d="M10 4 6 8l4 4"
+              stroke="currentColor"
+              stroke-width="2"
+              stroke-linecap="round"
+              stroke-linejoin="round"></path>
+          </svg>
```
Anca2022: The icons in the preview don't match the current Figma (has the design shifted since this was put together?)
Also, would it make sense to add an `<Icon>` component, instead of inlining the `<svg>` every time?

Image

src/components/shared/LinkButton.astro
```diff
@@ -0,0 +1,61 @@
...
+  : {}
+const resolvedHref = ariaDisabled ? undefined : href
+---
+
+<a
+  href={resolvedHref}
+  class={twMerge(
+    buttonVariants({ variant, mode, size, iconOnly, class: className })
+  )}
+  data-umami-event={umami}
```
Anca2022: For analytics, there's already a PR that auto-instruments `data-umami-event` on links (See PR #256) - flagging in case it slipped past you, since we're turning away from manually adding umami props. 
I'm not saying you should handle it in this PR, but if you don't, could you open a follow-up issue so it doesn't get lost? 😁

Anca2022 (commented):
Waiting on the final Figma, but in the meantime I left a few comments and flagged one bug.
Other than that, this truly looks great, Ravi! 😁

src/components/shared/buttonVariants.ts
Infi-Knight: Good catch, fixed in 0cfbea0b. Root cause: `:hover` still fires on `<button disabled>` natively (the spec only blocks click events, not pointer states). The compoundVariants for secondary now explicitly reset `bg`/`border`/`text` under both `disabled:` and `aria-disabled:`, so the disabled state is fully inert visually.

src/pages/preview/ui-components.astro
Infi-Knight: Refactored in 0cfbea0b: added [src/components/shared/Icon.astro](src/components/shared/Icon.astro) keyed by name (`chevron-left`, `chevron-right` for now), default `size-4` overridable via `class`, `currentColor` so it inherits from the parent button text colour. The preview page now uses `<Icon name="chevron-right" />` instead of inlined SVGs.

On the Figma mismatch: the chevrons in the preview page are placeholders: icons are slot content the caller passes in. Adding the exact Figma icon set is a separate piece of work (icon extract / sprite pipeline / naming convention), so I would prefer to keep this PR focused. Happy to open a follow-up issue once we agree on the approach. Does that work?

src/components/shared/LinkButton.astro
Infi-Knight: Removed the `umami` prop in 0cfbea0b. The wrapper was incomplete anyway (only set `data-umami-event`, not the full `data-event-*` set that `buildUmamiAttrs` produces). Consumers can pass `data-umami-event` (and any `data-event-*`) directly via the rest spread on `<LinkButton>`, or use `buildUmamiAttrs` and spread the result. PR #256 covers MDX inline links via the rehype plugin; component-level instrumentation stays at the call site where it has the right context.

src/styles/base/variables.css
Infi-Knight: remove reference

JoblersTune:
The 404 page seems to have broken
https://deploy-preview-266--interledger-org-v5.netlify.app/hello
image

JoblersTune:
My reading of Figma is that our Buttons and LinkButtons should behave the same. That means no underline on hover in either and `pointer: cursor` for both.

JoblersTune:
When I look at Figma the secondary buttons I see don't have a black fill on hover, only the outline color changes
image

JoblersTune:
I see `src/components/shared/Icon.astro` uses a chevron icon but from Figma I see arrow icons.
image

JoblersTune:
The focus state in Figma and this PR also differ
image
versus
image

Anca2022:
The focus state behaves as expected locally, but in the deployed preview the focus ring inherits the accent-color from @interledger/docs-design-system (`outline-color: var(--sl-color-accent)`)

image
image
image

src/components/shared/Icon.astro
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+const { name, class: className } = Astro.props
+const { viewBox, path } = ICONS[name]
+---
+
+<svg
+  class={twMerge('size-4', className)}
```
Anca2022: From what I can see in Figma, this icon is only using `size-5` (20×20). We’d need to update that and also remove the `class="size-5"` prop from the `<Icon>` component used in /preview/ui-components.astro.

Infi-Knight:
> The focus state behaves as expected locally, but in the deployed preview the focus ring inherits the accent-color from @interledger/docs-design-system (`outline-color: var(--sl-color-accent)`)
> 
> image image image

THIS leak is the reverse: docs CSS leaking INTO main-site, via a different shared-chunk path that wasn't touched. working on a fix

Infi-Knight:
Traced the leak end to end. Not specific to this PR. The same `index@_@astro.<hash>.css` chunk is loaded on staging's home page already, and contains the unscoped `:focus-visible { outline-color: var(--sl-color-accent); border-radius: var(--border-radius); }` rule from `@interledger/docs-design-system`. Astro middleware (auto-registered by Starlight) drags the docs `customCss` chunk onto every route, including main-site ones. The new button focus states are just the first prominent main-site focus styling, so the bleed only became visible now.

Two other universal-ish rules in the same file bleed too with less visible damage: `img { border-radius }` rounds every `` on main-site, and `thead tr:first-of-type th:first/last-of-type` rounds the top corners of every table.

Filed upstream: interledger/docs-design-system#71 — suggests wrapping the rules in `@layer ilf-docs.base` or scoping with `:where(.sl-container)`.

Shipped a `revert-layer` band-aid in `src/styles/tailwind.css` for this PR (commit 26a0c547). Covers all three leaked rules. Once #71 lands and we bump the dep, the workaround can be removed.

src/components/shared/Icon.astro
Infi-Knight: Done — defaulted `Icon.astro` to `size-5` and removed the redundant `class="size-5"` props from ui-components.astro (commit 26a0c547).

src/components/shared/buttonVariants.ts
```diff
@@ -0,0 +1,127 @@
...
+          'aria-disabled:border-neutral-50 aria-disabled:text-neutral-50'
+        ]
+      },
+      {
+        variant: 'secondary',
+        mode: 'dark',
+        class: [
+          'border-neutral-75 text-neutral-25',
+          // Hover: border colour only changes to white; bg and text stay put.
+          'hover:border-neutral-0',
```
Anca2022, line 99:
```suggestion
          'hover:border-neutral-0 hover:text-neutral-0',
```
nit, you can barely see the difference, but to be consistent with Figma

src/components/shared/buttonVariants.ts
```diff
@@ -0,0 +1,127 @@
...
+          // Focus: neutral-75 fill, 2px white inset, white text.
+          'focus-visible:bg-neutral-75 focus-visible:text-neutral-0',
+          'focus-visible:outline-2 focus-visible:outline-solid focus-visible:-outline-offset-2 focus-visible:outline-neutral-0',
+          'disabled:border-neutral-100 disabled:text-neutral-75',
+          'aria-disabled:border-neutral-100 aria-disabled:text-neutral-75'
+        ]
+      },
+      { iconOnly: false, size: 'lg', class: 'px-xl' },
+      { iconOnly: false, size: 'sm', class: 'px-md' },
```
Anca2022, line 108: The only spot I found that doesn't match Figma EXACTLY — and I think Ravi's implementation is the right call. 
Figma specifies three padding cases:
- iconOnly: square ✅
- text-only: symmetric x-padding ✅ 
- text + icon: falls back to the text-only symmetric padding; Figma calls for **asymmetric** — wider on the text side, tighter on the icon side

The asymmetric case adds real complexity for a few pixels of optical balance most users won't notice. Ravi's symmetric padding keeps the matrix clean; I'd ship as-is and revisit only if design pushes back.

Anca2022 (approved):
Everything looks great, Ravi! 🎉

PR #269, opened 2026-05-05
refactor(utils): split src/utils into shared/main/docs lanes [INTORG-697]

Infi-Knight:
## Summary

Reorganizes `src/utils/` into three lane buckets so the docs/main-site safety boundary is visible at import time:

- `src/utils/shared/`: pure helpers usable on either side of the boundary (currently just `url.ts`)
- `src/utils/main/`: routing, i18, summit, blog, content collections, navigation, analytics, and other main-site code
- `src/utils/docs/`: Starlight-only helpers (`rewriteRfcLinks`, `parseRawGitHubPath`)

Pure refactor. No logic changes.

The barrel `src/utils/index.ts` re-exports `shared/` and `main/` only. `docs/` is intentionally out of the barrel so docs consumers must go through `@/utils/docs/*` directly. Inside `src/utils/`, cross-file imports are relative (`./foo` within a lane, `../<lane>/foo` across lanes), matching the existing `cms/src/utils/` convention.

Also includes a small drive-by fix to `rehypeUmamiLinks.test.ts`: a typo'd `toContain` assertion (the expected string had the attribute name doubled) was failing on `staging` too. Verified on a clean `staging` worktree. Fixed in a separate commit. After the fix, all 53 root tests pass.

## Related Issue

Fixes INTORG-697.

Note for reviewers: this only makes the boundary visible. It does not enforce it. Without INTORG-654 (the dependency-cruiser boundary check, which is still open), nothing prevents a future `@/utils/main/*` import from sneaking back into `src/components/docs/`. The CSS-leak mechanism described in INTORG-639 still applies.

## Manual Test

- Build: `pnpm build` produces 1192 prerendered pages with no errors
- Lint: `pnpm lint` passes
- Tests: `pnpm test` passes (53/53) after the assertion fix

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [ ] Scope is focused (target ~10-20 files when possible). This is 91 files, but they are all mechanical import-path rewrites driven by the move; touching fewer would leave broken imports
- [ ] Screenshots for UI changes (N/A, no UI)

## Notable details

- Two **dynamic imports** in `extractSessionize.ts` (`../data/sessionize/${year}-*.json`) needed an extra `..` after the file moved one level deeper. Caught by the production build, fixed.
- `astro.config.mjs`, `scripts/optimize-images.ts`, `scripts/sync-sessionize.ts`, and `src/data/docs/rfcs.ts` also needed path updates since they reach into utils via either alias or relative paths.
- All file moves done via `git mv` so history is preserved.

## Follow-up worth considering separately

The root `pnpm test` is not wired into `pr-checks.yml` (only the CMS suite runs). That is why the assertion typo above sat undetected. Adding the root test job to the PR-checks workflow would be a 3-line change in a small follow-up PR.


CLAUDE.md
```diff
@@ -90,11 +90,22 @@
 
 ## Shared Utilities (`src/utils/`)
 
-- **Always check `src/utils/index.ts` before writing a new utility function** — it's the full catalog of what exists, grouped by domain. Scan the relevant module if you need more detail.
-- If a utility already exists, import it from `@/utils` — never duplicate it inline.
+`src/utils/` is split into three lane buckets so the docs/main-site CSS isolation (see `src/styles/README.md` "Starlight Docs Isolation") stays visible at import time:
```
Anca2022, line 93: The `src/styles/README.md` currently only covers docs vs main-website isolation from the CSS angle. 
Worth adding a short section on the JS-side counterpart: the three `shared`, `main` and`docs` buckets and why docs-side consumers should import via direct subpaths rather than the @/utils barrel.

Anca2022 (approved):
I pushed a commit that
- refactors imports through central `src/utils/index.ts` barrel instead of deep subpath imports (e.g. `@/utils/main/umami`)
   This aligns with the import conventions documented in CLAUDE.md
("Use the barrel @/utils for shared/ and main/ exports").
- updates some paths to util functions in the root README
- deletes an empty unused component (`/components/pages/Breadcrumbs.astro`) - for breadcrumbs we are using `components/shared/Breadcrumbs.astro` component

Made a comment about maybe adding some information about the docs - main website isolation from a JS point of vue in the README (styles README / root README), but that's not a blocker :)

PR #281, opened 2026-05-12
feat(homepage): replace content-driven home with dedicated hero compo… INTORG-658

JonathanMatthey:
## Summary

- Replaces the MDX content-collection approach for the homepage (`home.mdx` + `FoundationContentPage`) with a purpose-built `HomePage` component and a dedicated `HomeHero` section
- Hero section: full-viewport dark background, hero image (`stefan-thomas.webp`), fluid headline in two colour spans, body copy, and two CTA buttons using `LinkButton` (primary + secondary dark-mode)
- All visible strings extracted to `ui.ts` in both `en` and `es` — zero hardcoded copy in the component
- Umami `hero`-section tracking on both CTAs via `buildUmamiAttrs`; `data-component="HomeHero"` on the section root for automated analytics attribution

## Why

The old homepage was driven by a CMS content block (`FoundationContentPage` + `home.mdx`) which forced all layout decisions into generic block components. The new design requires a fixed hero structure that can't be expressed in that model cleanly — a dedicated component gives full layout and a11y control without bending the CMS abstractions.

## Accessibility & performance notes

- Hero `<section>` is labelled via `aria-labelledby` pointing to the `<h1>`
- Background image container is `aria-hidden="true"` with `alt=""` on the `` — purely decorative
- Image uses `loading="eager"` + `fetchpriority="high"` for LCP; gradient overlay is CSS-only (no extra resource)
- Buttons use the canonical `LinkButton` component which ships with WCAG-compliant focus states from `buttonVariants`

## Test plan

- [ ] `/` renders the hero in EN — headline, description, both CTAs visible
- [ ] `/es/` renders the hero in ES with Spanish copy
- [ ] Tab through the page: both CTA buttons show visible focus rings


Infi-Knight (approved):

src/components/pages/HomePage.astro
JoblersTune, line 1: Why is this created as a component? Surely this should go in index.astro directly? Can just the hero be broken out into a component?

src/data/ui.ts
```diff
@@ -327,6 +333,13 @@ export const ui = {
       'Gracias por contactarnos. Le responderemos lo antes posible.',
     'contact.error.captcha': 'Por favor, complete el desafío de reCAPTCHA.',
     'contact.error.submit':
-      'Hubo un problema al enviar su mensaje. Por favor, inténtelo de nuevo.'
+      'Hubo un problema al enviar su mensaje. Por favor, inténtelo de nuevo.',
```
JoblersTune, line 336: Please remove the Spanish UI translations here. We want them left blank intentionally.

src/components/pages/HomePage.astro
JoblersTune: Ah, this to allow the same page for EN and ES? That makes sense. 
Can we still break the hero out into a separate component though?

PR #289, opened 2026-05-13
Jm/homepage hero component

JonathanMatthey:
## Summary

- Extract homepage hero into a standalone `HomepageHero` component, decoupling it from `HomePage.astro`
- Add `media` prop to `OptimizedImage` so the Stefan Thomas background image is never fetched on mobile — the `<source media="(min-width: 1024px)">` attribute prevents the browser request entirely, rather than relying on CSS `display: none` which does not reliably suppress network fetches
- Clear Spanish translation values in `ui.ts` — all `es` keys are preserved as empty strings, falling back to English at runtime, ready for real copy when available

## Changes

**`src/components/pages/HomepageHero.astro`** (new)
Owns all hero logic: translations, Umami attrs, CTA hrefs, and the section markup. Accepts `pageLang` and `pathname` as props.

**`src/components/pages/HomePage.astro`**
Now a thin layout wrapper — just `FoundationPageLayout` + `<HomepageHero>`.

**`src/components/shared/OptimizedImage.astro`**
New optional `media` prop. When set, it is passed to every `<source>` element and the `` fallback `src` is omitted (`undefined`), so browsers that don't match the media condition have nothing to fetch.

**`src/data/ui.ts`**
All `es` locale string values replaced with `''`. Keys are intact; `useTranslations` already falls back to `en` for empty values.

## Test plan

- [ ] Desktop: hero image loads, `fetchpriority="high"` and `loading="eager"` are present in the rendered `<source>`
- [ ] Mobile (DevTools throttled): confirm no network request for `stefan-thomas.webp` in the waterfall
- [ ] Lighthouse Mobile: no unused image penalty
- [ ] Lighthouse Desktop: LCP is the hero image, score unaffected
- [ ] `/es/` routes: pages render using English fallback copy where `es` values are blank
- [ ] No TypeScript errors (`pnpm tsc --noEmit`)

image


src/components/pages/HomepageHero.astro
```diff
@@ -0,0 +1,91 @@
...
+    />
+  </div>
+
+  <div
+    class="mx-auto w-full max-w-wide px-space-m py-space-2xl md:py-space-3xl lg:min-h-[80vh] flex items-center"
+  >
+    <div class="max-w-136 md:max-w-144 lg:max-w-168">
+      <h1
+        id="home-hero-title"
+        class="font-poppins text-h2 md:text-h2-md lg:text-h2-lg"
```
Infi-Knight: since Poppins is now the default we should instead switch it at a global level to avoid repeatition

Infi-Knight (approved):

src/components/pages/HomepageHero.astro
JonathanMatthey: fixed :) thx

PR #293, opened 2026-05-14
Jm/improve optimized image to support avif

JonathanMatthey:
feat(images): add AVIF support to OptimizedImage pipeline

## Summary
- Encode `.avif` alongside `.webp` for every target width and `-full` variant in `optimize-images.ts` (q75, 4:4:4 chroma subsampling for cleaner dark gradients).
- Accept `.avif` as a source extension so masters can land as AVIF and still get full responsive variant generation.
- `getOptimizedImage()` now also returns `avifVariants` / `avifFullSrc`; existing webp fields preserved so `getHeroSectionStyle` (CSS background fallback) keeps working.
- `OptimizedImage.astro` emits `<source type="image/avif">` before `<source type="image/webp">` — AVIF-capable browsers pick AVIF, others fall through to webp, then to the bare ``.

## Why
Hero images with subtle dark gradients (e.g. homepage hero) showed banding/clipping under webp at q80. AVIF's 10-bit pipeline + 4:4:4 chroma preserves the smooth dark regions; comparable file size at the network level.

## Test plan
- [ ] `pnpm optimize:images` runs clean and produces both `.webp` and `.avif` variants under `public/img/optimized/**`.
- [ ] DevTools → Network on the homepage: hero request resolves to a `.avif` URL in modern browsers (Chrome/Edge/Firefox/Safari 16+).
- [ ] Force-disable AVIF (or test in older Safari) and confirm the `.webp` source is served instead.
- [ ] CSS background-image consumers (`getHeroSectionStyle`) still load the webp `fullSrc`.
- [ ] No regressions in `<OptimizedImage>` callers that pass `media=` (only the matching source should fetch).


scripts/optimize-images.ts
```diff
@@ -15,9 +15,19 @@ const getPublicAssetPath = (urlPath: string): string =>
 const OUTPUT_BASE = getPublicAssetPath(IMAGE_URL_PATHS.publicOptimized)
 
 const WEBP_QUALITY = 80
+// AVIF at q75 with 4:4:4 chroma subsampling is visually comparable to webp at
+// q80 but typically 20-30% smaller, with cleaner dark gradients (no banding).
```
Infi-Knight, line 19: I maybe wrong but with 4:4:4 (no chroma subsampling), AVIF can match or even exceed WebP q80 in size on noisy photography; the banding/gradient win is real but the size win is content-dependent. Just wanted to point out

JonathanMatthey:
made a cleaner new PR - been merged - closing this

PR #308, opened 2026-05-22
feat(home): add Lenis smooth scroll to homepage [INTORG-734]

Infi-Knight:
## Summary

Adds wheel/trackpad smooth scrolling to the homepage via [Lenis](https://github.com/darkroomengineering/lenis), mirroring the Framer prototype at [inspirational-lottery-436833.framer.app](https://inspirational-lottery-436833.framer.app/). Scoped to the homepage only via a `<SmoothScroll />` loader component — docs, blog, and other routes keep native scroll.

Implementation choices:

- **Lenis config**: `duration: 0.9` (everything else default). Pulled from the Framer prototype's compiled bundle — its Smooth Scroll component uses `new Lenis({ duration: intensity/10 })` and the prototype sets `intensity: 9`. Scroll-curve A/B against the live prototype settles to the same final position at the same time (~900ms, easeOutExpo).
- **Accessibility**: bails immediately when `prefers-reduced-motion: reduce` matches, and `destroy()`s if the preference flips at runtime. Lenis does not honor this preference by default.
- **Anchor links**: same-origin `a[href*="#"]` clicks are intercepted and scrolled via `lenis.scrollTo`, with the offset read from the target's `scroll-margin-top`, falling back to `.foundation-header.offsetHeight` so sticky-header overlap is avoided.
- **Nested scrollers**: walks the DOM once at init and tags `overflow: auto` elements with `data-lenis-prevent` so they keep native scroll — same approach the Framer component takes.
- **CSS**: 5 standard Lenis rules added under `@layer base` in `reset.css`. They're inert unless `<html>` has the `lenis` class.

## Related Issue

Fixes INTORG-734

## Manual Test

1. Load homepage on desktop with a trackpad or mouse wheel — scroll should feel eased, not snappy. Compare A/B against [the Framer prototype](https://inspirational-lottery-436833.framer.app/) — feel should be near-identical.
2. DevTools → Rendering → emulate `prefers-reduced-motion: reduce` → reload. `<html>` should not have the `lenis` class; scrolling should be native/instant.
3. Navigate to `/about-us`, `/blog`, or any docs page — scroll should be native (Lenis only loads on the homepage).
4. Tab through focusable elements — focus management is unchanged.
5. Mobile / touch — scroll stays native (Lenis's `syncTouch` default of `false`).

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused (6 files)
- [ ] Screenshots for UI changes (smooth scroll is motion-only; see scroll-curve comparison above)

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "dotenv": "^17.2.4",
     "gray-matter": "^4.0.3",
     "html-to-text": "^9.0.5",
+    "lenis": "^1.3.23",
     "markdown-it": "^14.1.0",
     "marked": "^17.0.4",
     "node-html-parser": "^7.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -55,6 +55,9 @@ importers:
       html-to-text:
         specifier: ^9.0.5
         version: 9.0.5
+      lenis:
+        specifier: ^1.3.23
+        version: 1.3.23(react@18.3.1)
       markdown-it:
         specifier: ^14.1.0
         version: 14.1.1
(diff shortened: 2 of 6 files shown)


JonathanMatthey (approved):
looks great

PR #316, opened 2026-05-27
Jm/intorg 759 improve lighthouse - INTORG-759

JonathanMatthey:
improve lighthouse


src/integrations/strip-docs-css-from-main-site.ts
Infi-Knight, line 1: is this a fix for https://linear.app/interledger/issue/INTORG-654/harden-docsmain-site-css-isolation-to-prevent-tailwind-leaking-into ?

Infi-Knight (commented):
SmoothScroll is not being used anywhwere, are we removing smooth scroll?

Infi-Knight (approved):

PR #318, opened 2026-06-01
feat(preview): add typography preset reference page [INTORG-726]

JonathanMatthey:
## Summary

Adds an internal typography reference page at `/preview/typography`

## Related Issue

Fixes INTORG-726

## Manual Test

- [ ] Open `/preview/typography` — page loads with Tailwind styles (Poppins, spacing, grid layout)
- [ ] Confirm three sections: Mobile, Tablet, Desktop — anchor nav jumps to each

## Checks

- [ ] `pnpm run format`
- [ ] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/src/components/preview/TypographyPresetRow.astro
+++ b/src/components/preview/TypographyPresetRow.astro
@@ -1,0 +1,35 @@
+---
+import { fontWeightLabel, TYPOGRAPHY_SAMPLE } from '@/data/typography-preview'
+import type { TypographyPreset } from '@/data/typography-preview'
+
+interface Props {
+  preset: TypographyPreset
+}
+
+const { preset } = Astro.props
+---
+
+<div
+  class="typography-preset-row grid grid-cols-1 border-b border-neutral-25 tablet:grid-cols-[240px_1fr]"
+>
+  <dl
+    class="grid grid-cols-[max-content_1fr] gap-x-lg gap-y-xs border-neutral-25 p-lg text-body-sm-standard text-neutral-75 tablet:border-r"
+  >
+    <dt>Font family</dt>
+    <dd>Poppins</dd>
+    <dt>Font style</dt>
+    <dd>{fontWeightLabel(preset.fontWeight)}</dd>
+    <dt>Font size</dt>
+    <dd>{preset.fontSizePx}</dd>
+    <dt>Line height</dt>
+    <dd>{preset.lineHeightPx}</dd>
+    <dt>Class</dt>
+    <dd>
+      <code class="text-neutral-100">{preset.className}</code>
+    </dd>
+  </dl>
+  <div class="p-lg">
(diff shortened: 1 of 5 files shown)


Infi-Knight (approved):

Infi-Knight:
### Drift risk: `typography-preview.ts` hand-copies values from `theme.css`

Reviewed the page and verified every preset against the `@theme` tokens — all 27 sizes/line-heights/weights are accurate today, and the breakpoint labels match `tailwind.config.mjs`. Nice work.

One thing worth flagging before merge: `src/data/typography-preview.ts` hardcodes px values that (per its own header comment) "mirror @theme tokens in `theme.css`." So each row has two independent sources of truth — the rendered sample is driven by the real CSS token via `className`, but the displayed `Font size: 56px` is a separate literal. The moment someone edits `theme.css`, this reference page silently goes stale and starts showing wrong numbers. For a design-system reference page specifically, a doc that can disagree with the system is worse than no doc.

They agree right now; nothing structurally keeps them in sync. Cheapest fix that fits our testing convention: a **parity test** that parses the `--text-*` / `--text-*--line-height` / `--text-*--font-weight` declarations out of `theme.css` and asserts each `TypographyPreset` matches — catches drift in CI, no runtime cost. (Alternatives: generate the catalog from `theme.css` at build time, or read `getComputedStyle` at runtime — though the latter adds client JS we generally avoid.)

Not blocking, but I'd lean toward the parity test before this becomes the canonical reference people trust.

JonathanMatthey:
@rav

> ### Drift risk: `typography-preview.ts` hand-copies values from `theme.css`
> Reviewed the page and verified every preset against the `@theme` tokens — all 27 sizes/line-heights/weights are accurate today, and the breakpoint labels match `tailwind.config.mjs`. Nice work.
> 
> One thing worth flagging before merge: `src/data/typography-preview.ts` hardcodes px values that (per its own header comment) "mirror @theme tokens in `theme.css`." So each row has two independent sources of truth — the rendered sample is driven by the real CSS token via `className`, but the displayed `Font size: 56px` is a separate literal. The moment someone edits `theme.css`, this reference page silently goes stale and starts showing wrong numbers. For a design-system reference page specifically, a doc that can disagree with the system is worse than no doc.
> 
> They agree right now; nothing structurally keeps them in sync. Cheapest fix that fits our testing convention: a **parity test** that parses the `--text-*` / `--text-*--line-height` / `--text-*--font-weight` declarations out of `theme.css` and asserts each `TypographyPreset` matches — catches drift in CI, no runtime cost. (Alternatives: generate the catalog from `theme.css` at build time, or read `getComputedStyle` at runtime — though the latter adds client JS we generally avoid.)
> 
> Not blocking, but I'd lean toward the parity test before this becomes the canonical reference people trust.

good idea...

done...

added a parity test file that will break CI if changes to theme dont match on this typography preview page

PR #326, opened 2026-06-03
feat(blog): add table block with table-only CKEditor preset [INTORG-780]

JonathanMatthey:
## Summary
- Style tables in foundation blog prose 
image

- Tables are inserted in the **Paragraph** block CKEditor toolbar 

Fixes INTORG-780

## Test plan
- [ ] Add a table to a Foundation Blog Post Paragraph block in Strapi


JonathanMatthey:
still working on mobile scroll - trickier than original

Infi-Knight:
Accessibility nit on the scroll wrapper:

The horizontal-scroll container is created purely in CSS via `& :has(> table) { overflow-x: auto }`, which targets the `Paragraph.astro` `<div set:html>`. A scrollable region needs to be reachable and operable by keyboard-only users — WCAG 2.1 SC 2.1.1 (Keyboard). The usual fix is to give the scroll container `tabindex="0"` plus an accessible name (e.g. `role="region"` + `aria-label`) so it can receive focus and be scrolled with arrow keys.

A `:has()` wrapper can't carry those attributes, so as-is a keyboard-only user can't scroll an overflowing table. Two options:

- Wrap the table in a real focusable element (e.g. have `Paragraph.astro` / the markdown renderer emit `<div role="region" aria-label="Table" tabindex="0">` around tables), then style that wrapper instead of `:has(> table)`, or
- Accept it as a known minor gap and track a follow-up.

Minor, but flagging since accessibility is non-negotiable per our conventions.

Infi-Knight (approved):

JonathanMatthey:
> Accessibility nit on the scroll wrapper:
> 
> The horizontal-scroll container is created purely in CSS via `& :has(> table) { overflow-x: auto }`, which targets the `Paragraph.astro` `<div set:html>`. A scrollable region needs to be reachable and operable by keyboard-only users — WCAG 2.1 SC 2.1.1 (Keyboard). The usual fix is to give the scroll container `tabindex="0"` plus an accessible name (e.g. `role="region"` + `aria-label`) so it can receive focus and be scrolled with arrow keys.
> 
> A `:has()` wrapper can't carry those attributes, so as-is a keyboard-only user can't scroll an overflowing table. Two options:
> 
> * Wrap the table in a real focusable element (e.g. have `Paragraph.astro` / the markdown renderer emit `<div role="region" aria-label="Table" tabindex="0">` around tables), then style that wrapper instead of `:has(> table)`, or
> * Accept it as a known minor gap and track a follow-up.
> 
> Minor, but flagging since accessibility is non-negotiable per our conventions.

so i have a solution to this, that lets you scroll using keyboard, tab then left / right keys... but... it's COMPLEX !!

image

PR #334, opened 2026-06-10
fix(cms): allow inline-code JSX in paragraph blocks [INTORG-793]

Infi-Knight:
## Summary

The paragraph-block save validator (`cms/src/utils/contentValidation.ts`) rejected any `<Capital…>` JSX-like tag in `blocks.paragraph` content unless it sat inside a **fenced** code block. It stripped fenced blocks before the scan but **not inline code spans**, so a tag written as literal inline code — e.g. `` `<wallet-address />` `` — was wrongly flagged and the save failed with a 400.

The developers blog uses inline-code tags throughout, so this blocks the foundation/tech blog merge (INTORG-691).

This adds `stripInlineCode()` (single- and multi-backtick spans) and runs it alongside the fenced-block strip before the JSX scan:

- ✅ inline-code JSX (`` `<MyComponent />` ``) — now allowed
- ✅ fenced-block JSX — still allowed
- 🚫 bare JSX outside any code — still rejected (existing protection preserved)

The error message now points authors to either inline backticks or a fenced block.

## Related Issue

Fixes INTORG-793. Unblocks INTORG-691.

## Manual Test

1. In Strapi admin, edit a blog post and add a **Paragraph** block.
2. In its content, include a tag as inline code, e.g. ``Use the `<wallet-address />` element.``
3. Save — it should save successfully (previously returned a 400 "JSX-like tag" error).
4. Sanity check the guardrail still holds: a bare `<Something>` tag (no backticks, not fenced) should still be rejected on save.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/cms/src/utils/contentValidation.ts
+++ b/cms/src/utils/contentValidation.ts
@@ -15,12 +15,25 @@ function stripFencedCodeBlocks(text: string): string {
 }
 
 /**
+ * Strip inline code spans (`…`, ``…``) so JSX/HTML tags written as literal
+ * inline code aren't mistaken for bare JSX. The developers blog routinely uses
+ * inline backticks to show tags like `<wallet-address />` in prose, so the
+ * merged blog must accept the same (INTORG-793).
+ *
+ * Runs after fenced blocks are removed; the remaining backtick runs are inline.
+ */
+function stripInlineCode(text: string): string {
+  return text.replace(/(`+)[\s\S]*?\1/g, '')
+}
+
+/**
  * Validate that no Paragraph block contains bare JSX-like tags.
  *
  * Returns a Strapi `ValidationError` when a `<CapitalLetter...` pattern is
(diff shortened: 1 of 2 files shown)


cms/src/utils/contentValidation.ts
```diff
@@ -35,13 +48,14 @@ export function validateNoNestedJsx(
       continue
     }
 
-    const stripped = stripFencedCodeBlocks(block.content)
+    const stripped = stripInlineCode(stripFencedCodeBlocks(block.content))
```
Anca2022:
```suggestion
    const stripped = stripInlineCode(block.content)
```
`stripInlineCode` replaces *any* number of backticks so it now does the job of `stripFencedCodeBlocks` redundant
Please consider deleting the dead code :)

Anca2022 (approved):
Looks good :D

PR #335, opened 2026-06-10
fix(cms): don't export blog article bios without an author [INTORG-794]

Infi-Knight:
## Summary

A blog post whose Strapi article-bio block has no author exported to MDX as `author: null`. Astro's content schema requires `articleBios[].author` to be a string (`ArticleBioSchema` in `src/schemas/content.ts`), so the post failed to build with `InvalidContentEntryDataError` and wouldn't render.

The two schemas disagreed: the Strapi `article-bio` component marked `author` as an optional string, while the Astro Zod schema requires it. Strapi saved the empty bio happily; Astro rejected the exported MDX.

This change:

- Makes `author` **required** on the Strapi `article-bio` component (`cms/src/components/shared/article-bio.json`) so an empty bio can't be saved in the first place.
- Skips authorless bios in the MDX export (`generateBlogMDX`) so existing/empty bios never emit `author: null`. When no valid bios remain, `articleBios` is omitted and the Astro schema falls back to its `[]` default.
- Exports `generateBlogMDX` and adds unit tests for null / empty / whitespace authors and mixed lists.

## Related Issue

Fixes INTORG-794. Surfaced while previewing a synced post during INTORG-691.

## Manual Test

1. In Strapi admin, edit a blog post, add an **Article Bio** block, and leave **author** empty. With this change the field is now required, so saving prompts you to fill it (previously it saved and produced a broken post).
2. For an already-synced post that exported with `author: null`, re-sync/re-export — the `articleBios` entry is dropped instead of writing `author: null`, and the post builds/renders.
3. A bio with a real author still exports normally (`- author: '…'`, optional `link`).

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

## Note for reviewers

Making `author` required is a Strapi content-type change. On environments with existing data, re-run `sync:mdx` / reset the dev DB as needed (same caveat noted for the INTORG-765 component changes). Existing bios all have authors, so no data migration is expected.

### DIFF
--- a/cms/src/components/shared/article-bio.json
+++ b/cms/src/components/shared/article-bio.json
@@ -6,7 +6,8 @@
   "options": {},
   "attributes": {
     "author": {
-      "type": "string"
+      "type": "string",
+      "required": true
     },
     "link": {
       "type": "string",
--- a/cms/src/index.ts
+++ b/cms/src/index.ts
@@ -5,7 +5,6 @@ import {
   scheduleGitSync,
   validateGitSyncRepoOnStartup,
   validateNoNestedJsx,
-  validateArticleBioAuthors,
   normalizeNavigationInput,
   LOCALES,
   shouldSkipMdxExport
@@ -1009,7 +1008,7 @@ export default {
         ctx: {
           method?: string
           url?: string
-          request?: { body?: { content?: unknown; articleBio?: unknown } }
+          request?: { body?: { content?: unknown } }
           status?: number
           body?: unknown
         },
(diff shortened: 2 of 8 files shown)


Infi-Knight:
Thanks for the careful review, @Anca2022

**Strapi not blocking empty required component fields:** this isn't a missing setting on our side, it's a known Strapi limitation. Strapi validates `required` on top-level fields but not on fields nested in components (especially repeatable ones): strapi/strapi#10030, strapi/strapi#14780. And because Strapi runs component create/update *before* the collection lifecycle hooks, lifecycle hooks can't catch it either. The only thing that can is a global content-manager middleware (the same pattern we use for `validateNoNestedJsx`).

I had added exactly such a middleware for the bio author, but per Sarah this broader validation gap is deprioritised for now, so I've backed it out to keep this PR focused. What stays here is the build-safety fix: the export skips authorless bios so they never emit `author: null` (so the site build can't break), and the component keeps `required: true` for the editor hint + types.

I've logged the real fix as a follow-up — **INTORG-797** (enforce required component fields via a reusable middleware, with a draft-vs-publish decision to make). So this is tracked, just not in this PR.

Anca2022 (approved):
LGTM 
Thanks, Ravi! 😄

PR #341, opened 2026-06-16
feat(blocks): add CTA strip component [INTORG-789]

Infi-Knight:
## Summary

Adds a CTA strip content block: a rounded banner with a heading, a description, a required primary CTA, an optional secondary (outline) CTA, and a background colour you can pick (purple or green, default purple). It uses the existing block pipeline, so it round-trips between Strapi and Astro.

**Strapi**
- New `blocks.cta-strip` component schema, with all content fields localized.
- Added to the `foundation-page` and `summit-page` dynamic zones and the populate config.
- Admin field labels and edit layout.
- MDX export serializer plus MDX-to-Strapi import handler. The handler validates the required fields, the colour enum, and that the secondary CTA has both its text and link, or neither.

**Astro**
- `CtaStrip.astro` reuses `LinkButton` (primary solid, secondary outline) with a trailing arrow icon and Umami `cta_button` analytics. The primary and secondary buttons are always the same width.
- Decorative Interledger network motif, recoloured per variant (periwinkle `#838FFF` strokes; on the green strip the largest node is emerald). It sits on the left and bleeds off the edge.
- Registered in both render paths: the Strapi JSON preview (`DynamicZone`) and the MDX component maps (`FoundationContentPage`, `SummitContentPage`). The description reads from a prop in the dynamic zone and from slot children in MDX.
- Added a showcase to `/preview/ui-components`.

It also fixes an older bug in the shared serializer escaping. `escDouble` and `escSingle` escaped non-ASCII characters to backslash-x sequences, which broke the round-trip for localized content. That fix is in its own commit.

## Related Issue

Fixes INTORG-789

## Manual Test

1. In Strapi, edit a Foundation Page, add a CTA Strip block, fill in the heading, description, and primary CTA, optionally add a secondary CTA, pick a colour, then Save.
2. Check the page exports to MDX with a `<CtaStrip ...>` block.
3. Open the page on the site and check it renders in both colours, with and without a secondary CTA, at mobile, tablet, and desktop widths.
4. Run `pnpm --dir cms sync:mdx:dry-run` and confirm it re-imports all content with 0 errors.

I tested this against a live Strapi: the block shows in the component picker, saves, exports to MDX, renders on the page, and re-imports cleanly. It is also on the `/preview/ui-components` page.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused
- [] Screenshots for UI changes.


cms/scripts/sync-mdx/ctaStripHandler.ts
```diff
@@ -0,0 +1,116 @@
...
+    ) {
+      throw new MdxParserError({
+        code: ParserErrorCode.CONFLICTING_PROPS,
+        message:
+          'CtaStrip secondary CTA requires both "secondaryButtonText" and "secondaryButtonLink", or neither.',
+        component: 'CtaStrip',
+        prop: 'secondaryButtonText',
+        line: node.position?.start.line,
+        column: node.position?.start.column
+      })
```
Anca2022: This error here (L71-79) doesn't throw for me, it is silently dropping Secondary button from the rendered component (when either text or link are missing)

Image

src/components/blocks/CtaStrip.astro
```diff
@@ -0,0 +1,139 @@
+---
+import { buildUmamiAttrs } from '@/utils'
+import Icon from '../shared/Icon.astro'
+import LinkButton from '../shared/LinkButton.astro'
```
Anca2022:
```suggestion
import Icon from '@/components/shared/Icon.astro'
import LinkButton from '@/components/shared/LinkButton.astro'
```
nit: for consistency

src/components/blocks/CtaStrip.astro
```diff
@@ -0,0 +1,139 @@
...
+  <div class="flex flex-col items-center gap-lg text-center">
+    <h2
+      class="font-poppins text-h3 tablet:text-h3-md desktop:text-h3-lg my-0 text-neutral-900"
+    >
+      {heading}
+    </h2>
+
+    {
+      hasDescription && (
+        <div class="font-poppins text-h4 tablet:text-h4-md desktop:text-h4-lg text-neutral-100 [&_p]:my-0">
```
Anca2022:
```suggestion
        <div class="font-poppins text-h4 [&_p]:text-h4 [&_a]:text-h4 tablet:text-h4-md [&_p]:tablet:text-h4-md [&_a]:tablet:text-h4-md desktop:text-h4-lg [&_p]:desktop:text-h4-lg [&_a]:desktop:text-h4-lg text-neutral-100 [&_p]:my-0">
```

Setting explicit text-sizes on descendant `p` and `a` or else the styles defined in `default.css` will apply here :) 
(which is fine in the context of this PR, but I've changes the default styles in another PR :D )

Anca2022 (commented):
Looking great, Ravi! :D 

Just an error that doesn't throw and some nits :)

cms/scripts/sync-mdx/cta-strip.serializer.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  it('escapes double quotes in attribute values', () => {
+    const result = serialize({
+      heading: 'The "best" offer',
+      description: 'Body.',
+      primaryButtonText: 'P',
+      primaryButtonLink: '/p'
+    })
+
+    expect(result).toContain('heading="The \\"best\\" offer"')
```
Anca2022: oh, and this test passes, but the page fails

Image

cms/scripts/sync-mdx/ctaStripHandler.ts
Infi-Knight: Good catch, the inconsistency was real. The `throw` only ran during the MDX import (the sync script), never in the Strapi -> render path, so in practice a half-filled secondary was just silently dropped. I made it all-or-nothing in every direction instead: the serializer now drops an incomplete secondary on export (the case you hit), the import handler drops it too rather than hard-throwing (one bad entry should not fail the whole sync), and the component already only renders it when both fields are present. Round-trip test added for the half-filled case. (10617131, af4b9584)

src/components/blocks/CtaStrip.astro
Infi-Knight: Done, applied your suggestion. (af4b9584)

src/components/blocks/CtaStrip.astro
Infi-Knight: Done, applied your suggestion. Pinned the descendant `p`/`a` sizes per breakpoint. (af4b9584)

cms/scripts/sync-mdx/cta-strip.serializer.test.ts
Infi-Knight: You were right. The serializer used jsesc, which emits a backslash escape (\") that is a parse error in a JSX attribute, so the page broke even though the test asserted that output. Switched the shared escapers to HTML-entity encoding (&amp;quot; / &amp;amp; / &amp;lt; / &amp;gt;); I confirmed the MDX parser decodes those back to the literal characters, so the export -> import round-trip holds. Added round-trip coverage for quotes/ampersands/angle brackets (the gap that let this through). The same latent bug was in the video-embed serializer, fixed too. (10617131)

Anca2022 (approved):
Looking great! 
Thanks, Ravi! 🎉

PR #215, opened 2026-04-11
feat(images): add image optimization pipeline with responsive WebP output INTORG-575

JonathanMatthey:
## Summary

Adds a pre-build image optimization pipeline using `sharp` that generates responsive WebP variants (640w, 1280w, 1920w) from raster images in `public/img/` and `public/uploads/img/original/`. 

Introduces an `OptimizedImage` component that emits `<picture>` elements with a WebP `<source>` and responsive `srcset`, falling back to the original `` when no variants exist. 

The component is wired into all MDX contexts via the `img` override.

## Related Issue

Fixes INTORG-575

## Manual Test

1. Run `pnpm run optimize:images` — verify `public/img/optimized/` is populated with WebP variants
2. Run `pnpm run build` — confirm `prebuild` runs optimization automatically
3. Open a page with inline MDX images and inspect the rendered HTML — should see `<picture>` with a `<source type="image/webp" srcset="...">` and multiple width entries

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] PR title follows Conventional Commits (e.g. `feat: ...`, `fix: ...`)
- [x] Linked issue included
- [x] Scope is focused (target ~10-20 files when possible)

PR #350, opened 2026-06-24
feat(roadmap): add SSR developers tech roadmap [INTORG-737]

Infi-Knight:
## Summary

Brings the public tech roadmap over from the standalone `interledger.org-developers` repo into v5, following Adi's merged PR #215. It is an SSR page at `/developers/roadmap` that reads a Linear snapshot from Netlify Blobs at request time. Two sync functions keep that snapshot fresh from the Linear API (details below), and the CDN caches the rendered HTML in between so in production the page itself never calls Linear.

- A timeline board grouped by team, with project duration bars, completed/extension bars, and milestone diamonds.
- Alternating row colors instead of Linear's per-project colors. This fixes the contrast issues Sarah raised on INTORG-636.
- A sticky header row, horizontal scrolling with a frozen left label column, and a quarter/full-timeline toggle on mobile.
- Umami analytics on milestone interactions.
- The timeline is capped to the current and next year, so older projects don't stretch the axis back years and leave rows looking empty.

Data wiring:

- The page reads the snapshot with `@netlify/blobs`. The `@linear/sdk` snapshot builder is ported from the source repo.
- `roadmap-sync` runs on a 12h cron. `roadmap-sync-now` is a manual `POST /api/roadmap-sync` (bearer auth, rate limited) for forcing an immediate refresh in deployed environments. Both fetch Linear, write the blob, and purge the CDN cache.
- Local dev needs no secrets and no manual sync. With no key set, the page renders a bundled fixture so it works with zero setup. Set `LINEAR_API_KEY` and the dev page fetches Linear directly under either `pnpm start` or `netlify dev` (the helper reads the key from both `process.env` and `import.meta.env`). The live snapshot is cached in-memory plus a 10-minute on-disk cache that survives HMR/restarts, so Linear is hit at most once per 10 min in dev. The env vars and the manual POST endpoint only matter for deployed environments; production shows an empty state until the blob is populated.

Caching:

- The page is served via Netlify Durable Cache (`Netlify-CDN-Cache-Control`, 12h), tagged `roadmap`. Durable Cache persists across deploys by design, so the data syncs purge the `roadmap` tag whenever they write the blob.
- Because Durable Cache survives deploys, a code/CSS-only change would otherwise keep serving stale HTML until the cache expired. The new `netlify/plugins/purge-roadmap` build plugin closes that gap: its `onSuccess` hook purges the `roadmap` tag on every deploy, so code changes show up immediately. It runs in the build context (auto-authenticated, deploy-scoped).

## Related Issue

Fixes INTORG-737

## Manual Test

- Fixture: `pnpm start`, open `/developers/roadmap`. The board renders with bundled fixture data, no secrets required.
- Live data: set `LINEAR_API_KEY` in `.env`, run `pnpm start` (or `netlify dev`), open `/developers/roadmap`. The page fetches Linear directly — no POST needed — and caches the result for 10 min so reloads/edits don't re-fetch.
- Cache purge on deploy: after a deploy, the Netlify build log shows `[purge-roadmap] purged roadmap CDN cache after deploy`, and the page reflects code changes without a manual sync.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [ ] Scope is focused
- [x] Screenshots for UI changes

Screenshot 2026-06-25 at 3 44 49 PM

Screenshot 2026-06-25 at 3 45 04 PM

Screenshot 2026-06-25 at 3 45 17 PM

### DIFF
--- a/netlify/functions/roadmap-sync-now.mts
+++ b/netlify/functions/roadmap-sync-now.mts
@@ -23,6 +23,13 @@ export default async function handler(
   req: Request,
   _ctx: Context
 ): Promise<Response> {
+  if (req.method !== 'POST') {
+    return new Response(JSON.stringify({ error: 'Method Not Allowed' }), {
+      status: 405,
+      headers: { Allow: 'POST', 'Content-Type': 'application/json' }
+    })
+  }
+
   const token = (req.headers.get('authorization') ?? '').replace(
     /^Bearer\s+/i,
     ''
--- a/netlify/functions/roadmap-sync-now.mts
+++ b/netlify/functions/roadmap-sync-now.mts
@@ -1,0 +1,79 @@
+import { timingSafeEqual } from 'node:crypto'
+import { getStore } from '@netlify/blobs'
+import type { Context } from '@netlify/functions'
+import { buildSnapshot } from '../../src/linear/build-snapshot'
+import { API_SECRET, isNetlifyDev } from '../../src/linear/env'
+import { purgeRoadmapCache } from './utils/purge-roadmap-cache.mts'
+
+const FIVE_MINUTES_MS = 5 * 60 * 1000
+const RL_KEY = 'sync-rate-limit'
+
+function unauthorized(): Response {
+  return new Response(JSON.stringify({ error: 'Unauthorized' }), {
+    status: 401,
+    headers: { 'Content-Type': 'application/json' }
+  })
+}
+
+// Manual sync trigger: POST /api/roadmap-sync with `Authorization: Bearer <API_SECRET>`.
+// Rate-limited to once per 5 minutes via the blob store.
+export default async function handler(
+  req: Request,
+  _ctx: Context
+): Promise<Response> {
+  const token = (req.headers.get('authorization') ?? '').replace(
+    /^Bearer\s+/i,
+    ''
+  )
+
+  if (!API_SECRET) return unauthorized()
+
+  const secretBuf = Buffer.from(API_SECRET)
+  const tokenBuf = Buffer.from(token)
+  const authorized =
+    tokenBuf.length === secretBuf.length && timingSafeEqual(tokenBuf, secretBuf)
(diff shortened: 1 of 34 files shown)


adrianboros:
Visually only the table bottom "Roadmap data refreshed at .... " it seems out of place, I would try to have the smallest possible font we use, darker color and align it with bottom left corner.

Infi-Knight:
Good catch, thanks Adrian. Restyled

- Smallest type token (`text-step--2`), darker (`text-gray-700`), left-aligned at the table's bottom-left.
- Also switched the timestamp from a fixed UTC string to the visitor's own locale and timezone. Since the page is SSR + CDN-cached the server can't know the viewer's locale, so it renders a UTC fallback (works without JS) and localizes on the client.

Should be visible on the preview once it redeploys.

netlify/functions/roadmap-sync-now.mts
```diff
@@ -0,0 +1,86 @@
...
+// Rate-limited to once per 5 minutes via the blob store.
+export default async function handler(
+  req: Request,
+  _ctx: Context
+): Promise<Response> {
+  const token = (req.headers.get('authorization') ?? '').replace(
+    /^Bearer\s+/i,
+    ''
+  )
+
```
adrianboros, line 37: Maybe add  also an early method guard:
something like: 

```
if (req.method !== 'POST') {
  return new Response(
    JSON.stringify({ error: 'Method Not Allowed' }),
    {
      status: 405,
      headers: {
        Allow: 'POST',
        'Content-Type': 'application/json'
      }
    }
  )
}
```

Infi-Knight: Done in ed356eb7. Added the early method guard at the top of the handler (returns 405 with an `Allow: POST` header before any auth or blob work). Thanks!

adrianboros (approved):

PR #369, opened 2026-07-04
fix(homepage): hero image jump fix -

adrianboros:
## Summary
On slower devices or weak connectivity there is jump of the hero image on the home page.
The jump happens because the image renders on page before the CSS custom properties (the transform/object-position) are calculated and applied. 
 - Repair the hero keyframe
 - Add paint containment to the hero wrap
 - refix positioning and bottom mask position

Secondary fix:
 - Eliminate nav CLS via synchronous inline script 
Here is what happens: HTML renders with data-offscreen="true". The external header-nav.ts module runs after the browser has already painted at least one frame. On desktop, the script immediately sets data-offscreen="false", triggering a CSS repaint. Total CLS is 0.0042 (well within "good"), but it contributes to the paint instability window.
- The fix: Add a small synchronous inline <script> block at the bottom of FoundationHeader.astro, placed after the nav wrapper element in the DOM. A synchronous script pauses HTML parsing and modifies the attribute before the browser paints that section, so the browser never paints the wrong state.

## Related Issue
INTORG-852

## Manual Test
Refresh the homepage, the hero image might have a slight delay to appear but should have the correct final position.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/src/components/layout/FoundationHeader.astro
+++ b/src/components/layout/FoundationHeader.astro
@@ -147,6 +147,20 @@ const headerClass = 'foundation-header sticky top-0 z-[2] px-lg desktop:px-2xl'
   </nav>
 </header>
 
+<!-- Synchronous init: set data-offscreen before first paint on desktop so the
+     browser never renders the drawer in mobile-closed state then flips it.
+     The module script below handles all interactions; this only sets the
+     initial attribute to match the viewport. -->
+<script is:inline>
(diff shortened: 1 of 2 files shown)


JonathanMatthey (approved):
very nice fix !

PR #370, opened 2026-07-06
feat(grant-faq): add grant faq component for both strapi and astro,

adrianboros:
## Summary
With this PR we add a new FAQ accordion section for the grant pages rendered on page between the main MDX body and the existing CTA Strip section.
The component has a two-column layout: a left column with title, subtitle, description, and a CTA button; a right column with accordion FAQ items. The design uses `<details>/<summary>` (zero-JS, natively accessible) with the existing `arrow-down` icon for the chevron.

## Related Issue
Fixes INTORG-846

## Manual Test
Open the `/grant/education/on-campus` page to view the rendered FAQ section with demo content

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


cms/src/components/blocks/grant-faq-item.json
```diff
@@ -0,0 +1,28 @@
...
+  "collectionName": "components_blocks_grant_faq_items",
+  "info": {
+    "displayName": "Grant FAQ Item",
+    "icon": "question-circle"
+  },
+  "options": {},
+  "attributes": {
+    "question": {
+      "type": "string",
+      "required": true,
```
JonathanMatthey, line 11: here marked as required true, but above they're optional 

FaqItem and FaqSection interfaces use optional fields (?), but the Strapi schema marks them all required: true. In lifecycles.ts the interfaces have question?: string, answer?: string, title?: string, etc. — all optional.

adrianboros: A good question, lets detail why this is happening..
So here is the Strapi schema definition which is more of UI/database validation/constraint.
The reason for having all fields optional in the lifecycle is the **trust boundary**: this lifecycle file receives data from Strapi's REST API at runtime, not from TypeScript's compiler. Even though the Strapi schema marks fields like `question` or `answer` as required, TypeScript has no way to verify that at compile time — the API payload arrives as an untyped JSON blob that gets cast (..../grant-page/lifecycles.ts line 56: page as GrantPageData).

Making the fields optional is the honest representation of what the code actually knows at the point it runs:

1. `as` **casts don't validate — they suppress.** Casting to an interface with required fields would silence the compiler without adding any real safety. If Strapi returned a null `question` despite it being "required", TypeScript would never warn you, and the bug would surface as a silent empty string or a crash.
2. **Optional fields force explicit handling.** Because `question` is `string | undefined`, the code at lines 97–100 is forced to write `i.question ?? ''` and `i.answer ?? ''.` That ?? is the actual guard — it makes the fallback visible and deliberate. With required fields you could write `i.question` directly and the compiler would stay quiet even when the value is missing.
3. The validators are the real enforcement. The validate callback on line 142–143 (`validateGrantPagePrimaryCta`, `validateGrantPageFaqSection`) is where the "required" constraint is actually enforced at runtime — before the MDX is generated. The lifecycle blocks the save if those fields are missing. The interfaces here describe the shape the code receives, not the contract it promises.

In short: Strapi's "required" is a UI/database-level constraint. These TypeScript interfaces describe what arrives over the wire — which could theoretically be anything — so ? is accurate, and the ?? '' fallbacks and validators are what substitute for the missing compile-time guarantees.

JonathanMatthey: 👌

JonathanMatthey:
design looks great, only thing is body paragraph text of accordion doesnt go full width, 

- [ ]  needs spacing on the right below the carets

figma:
image

yours:
image

- [ ] also look at font sizes

- [ ]  maybe a sliding css animation ?? rather than just open/close ?

JonathanMatthey:
- [ ] top of left heading and right heading - same horizontal start 

image

JonathanMatthey:
also please check with Radu, i have a feeling that on mobile `have more questions`. block disappears... 

your version
image

figma:
image



because it doesnt make sense to have it below the heading and below the questions, breaks the flow

JonathanMatthey (approved):
looks great - nice job ! :)

JonathanMatthey:
perfect 
image

PR #376, opened 2026-07-08
feat(split-layout): add split layout block for grant pages [INTORG-848]

JonathanMatthey:
## Summary

Adds a `blocks.split-layout` Strapi component and Astro block — two-column layout with image or video on one side, text or quote on the other. Includes a custom Strapi admin picker (4 icon buttons) that shows/hides the relevant fields per layout type. Wires up the full dynamic zone on `GrantPage.astro` and `GrantOverviewPage.astro`.

Also fixes a handful of pre-existing TypeScript / import-path issues that were blocking `strapi build`.

## Related Issue

Fixes INTORG-848

## Manual Test

1. In Strapi admin, open a Grant or Grant Overview page and add a Split Layout block
2. Confirm the 4 layout type buttons appear and selecting each shows only the relevant fields (image vs video, text vs quote)
3. Confirm "Image Position" label changes to "Video Position" when a video type is selected
4. Publish and verify the block renders correctly on the Astro frontend

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


Infi-Knight (commented):
left some comments

Infi-Knight:
## Review

Ran `tsc`, `eslint`, and vitest on `9d26f96` against an `origin/staging` baseline. This branch breaks grant MDX export, and it breaks it quietly.

### 1. `serializeContent` isn't exported from the utils barrel (blocker)

`grant-page/lifecycles.ts:11` and `grant-overview-page/lifecycles.ts:13` import it from `'../../../../utils'`. It lives in `cms/src/serializers/blocks/index.ts:43` and the barrel never re-exports it. `foundation-page/lifecycles.ts:7` gets this right.

```
lifecycles.ts(13,3): error TS2305: Module '"../../../../utils"' has no exported member 'serializeContent'.
lifecycles.ts(11,3): error TS2305: Module '"../../../../utils"' has no exported member 'serializeContent'.
```

Clean on staging. At runtime the binding is `undefined`, and `generateGrantPageMDX` calls it unconditionally, even when `content` is empty. `writeMDXFile` (`pageLifecycle.ts:255`) catches the `TypeError`, `exportAllLocales` (`pageLifecycle.ts:337`) logs it and continues.

Net effect: saving any grant or grant overview page silently stops writing its MDX file. No admin error, just a server log line. Pages without a split-layout block are hit too, so it regresses existing behaviour. Manual Test step 4 can't have passed.

Import from `'../../../../serializers/blocks'`, or export it from the barrel.

### 2. `pnpm run lint` fails 

```
cms/src/admin/app.tsx
  39:11  error  'CKEditor' is defined but never used      @typescript-eslint/no-unused-vars
  87:52  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  89:47  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
```

`editor: CKEditor` to `editor: any` orphaned the `CKEditor` interface and the four reachable only through it. Reconcile the types rather than casting.

### 3. Split layout enabled on `foundation-page`, nothing renders it (blocker)

`foundation-page/schema.json` and `FOUNDATION_PAGE_BLOCKS` both gain `blocks.split-layout`, and `pageLifecycle.ts:14` imports `serializeContent` correctly, so this path writes `<SplitLayout ...>` into foundation page MDX. `FoundationContentPage.astro:106` has no `SplitLayout` in its components map. On staging that map is exactly 1:1 with the dynamic zone; this makes it 9 components and 8 entries.

### 4. `layoutType` is required but only the admin UI reads it 

Nothing in `split-layout.serializer.ts`, `SplitLayout.astro`, or `splitLayoutHandler.ts` reads it. The picker hides fields with `display: none`, which leaves their values intact and published. Fill in Image + Text, switch to Video + Quote, and the stale `image` and `content` survive: the serializer emits `imageSrc`, `videoUrl`, `quote`, and a body. `SplitLayout.astro:73` takes the quote branch and drops `<slot />`, the heading, and the CTA.

Clear the fields on switch, or make `layoutType` authoritative. Right now it's neither.

### 5. `imageAlt` has nowhere to land

`splitLayoutHandler.ts` sets it and `types.blocks.ts` declares it, but `split-layout.json` has no such attribute. The serializer reads `alternativeText` off the media entity, and `resolveMediaUpload` is `(url) => Promise<number>` with no way to write it back. `blocks.image-block` uses a dedicated `altText`. The round trip loses alt text.

### 6. Grant-overview hero code is dead, and the delete loop is a landmine

`grant-overview-page/schema.json` has no `hero` attribute, and Strapi silently skips populate keys that don't match one (`@strapi/database` `populate/process.js:70-72`). So the populate entry, the `heroFrontmatter` spread, and the new `validate` hook do nothing.

`lifecycles.ts:50-58` is the problem: it unconditionally deletes `heroTitle`, `heroDescription`, `heroImage`, `heroImageAlt`, `heroImageMobile`, `heroImageMobileAlt`, `heroCtas` from `restPreserved` and re-adds nothing. `digital-finance.mdx` has no hero frontmatter today, so nothing breaks yet. Any page that gains one loses it on the next save. Looks like copy-paste from `foundation-page`. Remove it, or give grant-overview a real hero.

### 7. No tests

`COMPONENT_IMPORT_TEMPLATE.md` requires `<name>Handler.test.ts`. 

### Smaller things

- The picker finds fields by `gridRow.children[0]`, `previousElementSibling`, and `nextElementSibling`, all pinned to the row layout at `cms/src/index.ts:1080`. It also writes `textContent` onto a React-owned node, races the form with `setTimeout(..., 80)`, never re-runs when a dynamic zone block is reordered (which shifts every `content.N` prefix), and never cleans up on unmount. Select by `[name]` at minimum.
- `cms/tsconfig.json` adds `dom` and `jsx` across all of `cms/src`, so a lifecycle can reference `document` and still typecheck. Scope it to `src/admin/**` and `src/plugins/**`.
- Picker a11y: four exclusive options as `aria-pressed` toggles should be a radio group. No `:focus-visible`, no `aria-describedby` on `error`/`hint`, no visible label. Hardcoded hex (`#FFFFFF`, `#32324D`) will render white on white in Strapi's dark theme.
- `SplitLayout.astro` reimplements `Blockquote.astro` with different borders, a `<footer>` instead of `<cite>`, and no markdown on `quoteSource`.
- Media is always first in the DOM, flipped only at `tablet:` via `order-2`, so screen readers hit the image before the heading on the default `imagePosition="right"`.
- The `content` prop on `SplitLayout.astro` is never produced; the serializer emits children into `<slot />`. Set both and the body renders twice.
- `strapi-server.js` registers `type: 'string'` while the picker sends `type: 'text'`, and nothing validates the value is one of the four types.

JonathanMatthey:
addressed your feedback @Infi-Knight , some was WIP, i've cleaned up commits too

Infi-Knight (approved):
One minor note: `tsconfig.admin.json` exists but no package.json script points at it, so the picker and plugin `.tsx` are not typechecked by any command. `strapi build` compiles them, but a standalone `tsc -p tsconfig.admin.json` in the lint step would catch type errors earlier.

Nothing here blocks merge. Video handling stays as a plain `videoUrl` string, which is correct given uploads are tracked separately in INTORG-876.

PR #377, opened 2026-07-09
fix: failed validation clears unsaved content

adrianboros:
## Summary
In the Strapi admin, saving a page with a validation error would show the error but also wipe the editor's unsaved changes back to the last-saved DB state.

Root causes found (three separate, but related bugs)

1. Client-side cache bug in `@strapi/content-manager` — `updateDocument`'s RTK Query mutation optimistically writes edited data into the cache, then rolls it back (`patchResult.undo()`) on any error, and both `updateDocument`/`publishDocument` invalidated their cache tags even on failure. The admin's `<Form>` re-syncs to that reverted cache, wiping in-progress edits.
2. Custom validators ran too late — first fix moved validation into `beforeCreate`/`beforeUpdate` lifecycle hooks, but those see component fields already resolved to `{id, __pivot} `DB references. Initially moved validation to Koa middleware matching content-manager URLs — but that only covers the admin UI. The `sync-mdx`/`sync-navigation ` scripts write the same content types via the public REST API (used in CI on every staging/playground push) and were bypassing validation entirely. Replaced the Koa/URL-regex approach with `strapi.documents.use()` — a Strapi document-service middleware that runs below both the content-manager and public REST API controllers, so one registration now covers every write path.
3. Missing field-level error data — validators didn't attach `details.errors[].path`, and even after fixing that, the middleware's response body construction dropped `.details` entirely, so the admin could only show a generic toast, never highlight the specific field.

Fixes added:

- `patches/@strapi__content-manager@5.41.1.patch` (pnpm native patch, scoped in `cms/pnpm-workspace.yaml `— cms has its own standalone lockfile separate from the root workspace, which took a few tries to discover): stops the optimistic-update rollback and skips cache invalidation on error for updateDocument/publishDocument.
- `cms/src/utils/contentValidation.ts`: every validator (`validateGrantPagePrimaryCta`, `validateGrantPageFaqSection`, `validateCtaStrip`, `validateProfileCta`, `validateHeroFields`, `validateBlogFields`, `validateNavigationLabels`) now collects all failing fields per save (not just the first) via `combineFieldErrors`, and `mergeValidationErrors` combines multiple validators' results for one content type.
- `cms/src/index.ts`: `registerBodyValidationMiddleware` helper validates the raw request body pre-write and now correctly forwards details. Added missing validation entirely for `grant-overview-page` and `profile-page` (had none), and added `ctaStrip` validation to `grant-page`/`grant-overview-page `(a required component whose required sub-fields were never enforced on update).
- Reverted the now-wrong `beforeCreate`/`beforeUpdate` lifecycle validation additions in `pageLifecycle.ts`, `navigationLifecycle.ts`, `blogLifecycle.ts.`
- updated PR to include new feature merged into staging - grant page info-cards - validation wiring fix
- improved `mergeValidationErrors`
## Related Issue
Fixes INTORG-796

## Manual Test
- in a terminal to to cms/ folder and delete the node_modules/ folder and do a fresh `pnpm i`.
- Test validation on any page see if the issue is fixed and we don't lose content.
Check valid updates, check mdx file sync logic both ways.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
Screenshot 2026-07-09 at 23 39 57
Screenshot 2026-07-09 at 23 40 09
Screenshot 2026-07-09 at 23 40 26
Screenshot 2026-07-09 at 23 40 44


Anca2022:
There is still an issue with updating the entry - it is still deleting the content of a component not passing validation

**Nvm, this required node v24 to install the patch.**

https://github.com/user-attachments/assets/46a7cb94-1695-4fa4-8f92-b42c74a924ff

cms/src/utils/contentValidation.ts
```diff
@@ -20,6 +19,57 @@ export function toValidationError(error: unknown): errors.ValidationError {
...
+}
+
+/**
+ * Combine every field-level failure found by a single validator into one
+ * `ValidationError`, shaped like Strapi's own Yup validation errors
+ * (`details.errors: [{ path, message }]`). The admin's
+ * `_unstableFormatValidationErrors` iterates that array and highlights every
+ * offending field — so collecting all failures here (rather than stopping at
+ * the first) lets an editor fix everything in one pass instead of one save
+ * attempt per mistake.
```
Anca2022, line 34: 🔥

Anca2022 (approved):
This looks great! 🚀 

2 nice-to-have observations (don't know if they're possible):
1. can you join the validations errors that come from native Strapi with the ones we're enforcing?
2. can you also pass the 'path' for errors to components inside the dynamicZone?

No blocking though :D

PR #380, opened 2026-07-11
fix(cms): deleting a translated page didn't delete its MDX file [INTORG-869]

JonathanMatthey:
## Summary

Deleting a Spanish (non-English) entry in Strapi left its MDX file on disk. Both lifecycle delete hooks computed the file path from the deleted entry's own `pathSlug`, but non-English MDX files are actually named after the English sibling's slug — so whenever the two diverged, the path never matched and the delete silently no-opped. English deletes worked by coincidence since both locales share that same slug.

Fixed by resolving the English sibling's slug before computing the delete path for non-English entries, and deleting only that one locale's file (leaving the English source untouched).

## Related Issue
Fixes INTORG-869

## Manual Test
- Create an EN page + ES translation with a different `pathSlug`, publish both, delete only the ES entry in Strapi — confirm the ES `.mdx` file is removed and the EN file/entry are untouched.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/cms/src/utils/flatContentLifecycle.ts
+++ b/cms/src/utils/flatContentLifecycle.ts
@@ -264,10 +264,7 @@ export function createFlatLocaleMdxLifecycle<
           (loc) => getBaseDir(loc),
           label
         )
-        deleteLocaleMdxFiles(
-          (loc) => getFilePath(loc, result.pathSlug),
-          label
-        )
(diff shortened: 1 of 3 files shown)


adrianboros (approved):
tested locally, works

PR #384, opened 2026-07-14
feat(faq-template): adds FAQ page template base logic both Astro and Strapi

adrianboros:
## Summary
Adds FAQ as a new template type per [ADR-003]docs/decisions/003-reusable-page-templates-and-routing.md) — FAQ pages that can live under Foundation, Summit, or Hackathon via a section field, with no fixed URL prefix. Includes one example faq page and full round-trip sync support (Astro MDX ↔ Strapi).

## Related Issue
Fixes INTORG-865

## Manual Test
run mdx sync script
open the website at demo page url `/grant/faq` to view the bare bone demo page

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


src/lib/templates.ts
```diff
@@ -9,7 +9,8 @@ import type { CollectionKey } from 'astro:content'
  * No changes needed to getStaticPaths or existing collections.
  */
 export const crossSectionCollections = [
-  'profiles'
+  'profiles',
+  'faqs'
```
Infi-Knight: Adding `faqs` here means `getCrossSectionPaths` now emits `kind: 'faq'` entries for every section, including hackathon. But `src/pages/hackathon/[...page].astro` and `src/pages/es/hackathon/[...page].astro` weren't updated: they still render `<ProfileDetailPage>` unconditionally for whatever `getCrossSectionPaths('hackathon', ...)` returns.

So the first FAQ someone files under `section: hackathon` gets a static path generated, handed to ProfileDetailPage, which does `allProfiles.find(...)` and then `throw new Error('Profile not found: ...')`. That's a failed `astro build`, not a 404.

It isn't hypothetical either. The Strapi enum offers `hackathon`, and the admin hint you added in `cms/src/index.ts` tells editors to pick it when the FAQ sits under that microsite.

The root, `/es`, and both summit catch-alls all got the `kind === 'faq'` branch, so this looks like a straightforward miss. Both hackathon routes need the same branch. Since that branch is now copy-pasted across four route files, it might be worth pulling it into a small dispatcher component so template number three doesn't have to remember to touch five files.

cms/src/api/faq/content-types/faq/schema.json
```diff
@@ -0,0 +1,92 @@
...
+    "pathSlug": {
+      "type": "string",
+      "required": true,
+      "unique": true,
+      "pluginOptions": {
+        "i18n": {
+          "localized": true
+        }
+      },
+      "description": "Full path from the site root, no leading slash. Determines the public URL only. Example: grant/education/on-campus/faq."
```
Infi-Knight: This description is only true for `section: foundation`.

`src/pages/summit/[...page].astro` mounts cross-section paths under `/summit/` without stripping anything, and `buildSectionEntryBreadcrumbs` prepends the section again on top of `pathSlug`. So an editor who picks `section: summit`, reads "full path from the site root", and writes `pathSlug: summit/2025/faq` ends up at `/summit/summit/2025/faq`.

Worth saying the slug is relative to the section you pick, and that the `summit/` or `hackathon/` prefix must be left off. Same wording applies to the `pathSlug` hint in `cms/src/index.ts`.

(`profile-page`'s schema.json has the same misleading text, with summit examples no less. But every profile in the repo today is `foundation`, so nobody's hit it yet. No reason to carry it forward into the new type.)

src/utils/main/breadcrumbs.ts
```diff
@@ -0,0 +1,49 @@
...
+ * are routed.
+ *
+ * `urlLocale` must be the locale of the URL being rendered (e.g.
+ * `Astro.locals.routeLocale`), not the content locale. On a localized route
+ * that falls back to EN content (`isFallback`), the content locale is `en`
+ * but the page still lives under the localized URL prefix (e.g. `/es/...`) —
+ * using the content locale here would generate non-localized breadcrumb
+ * hrefs and bounce the visitor out of the localized section.
+ */
+export function buildSectionEntryBreadcrumbs(
```
Infi-Knight, line 29: The docstring above spells out exactly why using the content locale for breadcrumb hrefs is wrong: on an ES page that falls back to EN content, you generate non-localized hrefs and bounce the visitor out of `/es/`.

That is precisely what `ProfileDetailPage.astro` still does, inline, around lines 65 to 86. It builds the same Home / section / parents / label trail by hand with `localizeRoute(..., contentLocale)`.

So the helper was extracted and then only wired into the new page, leaving the buggy copy behind on the page it was extracted from. Swapping ProfileDetailPage over deletes about 18 lines and fixes ES-fallback profile breadcrumbs at the same time.

Smaller thing while you're in here: `BreadcrumbItem` on line 5 is now the second declaration of that shape. `Breadcrumbs.astro` declares an identical local interface. Worth importing this one there.

src/components/pages/FaqPage.astro
```diff
@@ -0,0 +1,78 @@
...
+  description={description}
+  headerTheme="light"
+>
+  <main
+    class="flex flex-col px-lg pt-[calc(var(--spacing-5xl)-var(--site-header-height))] pb-5xl tablet:px-xl tablet:pt-[calc(var(--spacing-6xl)-var(--site-header-height))] tablet:pb-6xl desktop:px-5xl desktop:pt-[calc(var(--spacing-7xl)-var(--site-header-height-desktop))] desktop:pb-7xl"
+  >
+    <TranslationDisclaimer isVisible={isFallback} />
+    <div class="max-w-content">
+      <Breadcrumbs breadcrumbs={breadcrumbs} />
+      <h2
```
Infi-Knight: This is the page's top heading, but it's an `h2`, and `FoundationPageLayout` doesn't contribute an `h1` of its own. So FAQ pages ship with no `h1` at all, which hurts both screen reader navigation and SEO.

`FoundationContentPage`, `GrantPage` and `GrantOverviewPage` all use `h1` for this. ProfileDetailPage uses `h2`, but that's a bug to fix rather than a pattern to copy.

The `text-h2` utility classes can stay if you want the same visual size; it's the element that needs to change.

cms/scripts/sync-mdx/config.ts
```diff
@@ -152,6 +155,29 @@ export function buildContentTypes(
         )
       }
     },
+    faqs: {
+      dir: getContentPath(projectRoot, 'faqs'),
+      apiId: 'faqs',
+      schema: faqFrontmatterSchema,
+      buildPayload: (mdx, strapi, existing, _dryRun) => {
```
Infi-Knight: Two things here.

First, this is the third verbatim copy of the `resolveMediaUpload` closure in this file (the others are in `buildPagePayload` around line 94 and in `profiles` around line 140). It wants to be a `createMediaUploadResolver(strapi)` helper.

Second, and more to the point: `buildFaqPayload`'s docstring says FAQ pages need "no media or relation resolution", yet this hands it both `resolveRelation` and `resolveMediaUpload`. One of the two is wrong. Given the FAQ dynamic zone only accepts `blocks.paragraph`, neither resolver can ever fire, so I'd just drop them and keep the docstring honest.

Same story with the ignored `_dryRun`: harmless today because paragraph blocks never resolve media, but it's a quiet trap for whoever widens the dynamic zone later.

Infi-Knight (commented):
Nice work on this. The content type mirrors `profile-page` closely, `buildFaqPayload` comes with real tests, and generalising `getProfilePaths` into `getSectionFilteredPaths` with an exhaustive `never` switch is the right call.

My comments are almost all on the routing side rather than the Strapi side. One of them (the hackathon routes) will break the build as soon as an editor uses a section the schema explicitly offers them, so I'd want that sorted before merge. The rest are cleanups.

One thing not worth an inline comment: `buildSectionEntryBreadcrumbs` has no tests. There are 16 test files in `src/utils/main/` and the function is pure, so it's cheap to cover: foundation (no prefix), summit (prefix added), a single-segment pathSlug (no parent crumbs), and the ES href case its own docstring is about.

Infi-Knight (commented):
Re-reviewed after the `fix part two` and lint commits. Everything from the last pass is handled, and cleanly:

- The hackathon routes (en and es) now branch on `kind` before rendering, so a `section: hackathon` FAQ no longer takes down the build. That was the one blocker.
- The `pathSlug` field description got the fix in both `schema.json` and the admin hint in `cms/src/index.ts`: it now says the slug is relative to the chosen section and that the `summit/`/`hackathon/` prefix comes off. Nice that you gave each section its own worked example rather than one generic line.
- `ProfileDetailPage` now calls `buildSectionEntryBreadcrumbs` with `routeLocale` instead of hand-rolling the trail with the content locale, so ES-fallback profile breadcrumbs are fixed too and the duplication is gone.
- FAQ pages render an `h1` now.
- The FAQ config entry drops both resolvers with a comment explaining why, and `buildFaqPayload`'s docstring matches reality again.
- `buildSectionEntryBreadcrumbs` picked up tests, including the ES-fallback href case its docstring describes, and `Breadcrumbs.astro` imports the shared `BreadcrumbItem` type.

Ran the touched suites locally: the 5 breadcrumb tests and 6 `buildFaqPayload` tests pass. No open concerns left from me.

Infi-Knight (approved):

PR #388, opened 2026-07-15
fix(hero): slow image parallax to trail the frame INTORG-895

JonathanMatthey:
## Summary
- `PageHero`'s parallax had it backwards — the image moved faster than the frame instead of slower. 
- `PARALLAX_SPEED` (currently 0.75, tunable 0.7–0.8) of the frame's scroll delta, clamped to the slack that height gives it. Also dropped the now-unused `simple-parallax-js` dependency.

## Related Issue
Fixes INTORG-895

## Manual Test
- Load a page using `PageHero` (e.g. a grant overview page) and scroll past the hero.
- Confirm the image lags behind the frame rather than outrunning it.
- Confirm no gap appears above/below the image while scrolling.
- Enable "reduce motion" in OS settings and confirm the image stays static.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


src/components/shared/PageHero.astro
```diff
@@ -113,27 +117,59 @@ const {
...
+      const rect = wrapper.getBoundingClientRect()
+      const totalTransit = window.innerHeight + rect.height
+      const progress = Math.max(
+        0,
+        Math.min(1, (window.innerHeight - rect.top) / totalTransit)
+      )
+      // Frame's slack is 30% of its own height (image is 130% tall),
+      // split evenly above/below — this is the max the image can trail.
+      const maxOffset = (rect.height * 0.3) / 2
+      const offset = maxOffset * (1 - 2 * progress)
```
Infi-Knight: the PR's purpose is "make the image move slower than the frame (trail it)." but I feel like the image moves faster than the frame's speed. The mechanism was rewritten, but the direction I think is still backwards.

So if we reverse the math

```
const offset = maxOffset * (1 - 2 * progress)   // current: 1.123× → outruns
const offset = maxOffset * (2 * progress - 1)   // suggestion:   0.877× → trails
```

seems to produce the intended trailing effect  and still fully covers the frame at every scroll position. Let me know what you think

public/uploads/img/original/DMW_Dno_Cnt6p_C70pha_SOX_0o_Aqu_Y_f819cc7b3d.webp
adrianboros, line 1: This file is huge, imo we should have something smaller in size.

JoblersTune: We should have a limit on image upload sizes. I would cap images at 2MB. I'll make an issue.

JoblersTune: https://linear.app/interledger/issue/INTORG-904/image-uploads-should-be-capped-at-2mg

JonathanMatthey: converted to avif, but a 2mb limit on a 4000px/3000px retina quality isn't realistic.

src/components/shared/PageHero.astro
JonathanMatthey: Radu is happy with the animation as is but i think you're correct.  updated - let me know

Infi-Knight (approved):

PR #420, opened 2026-07-23
feat(split-layout): add display ratio options [INTORG-937]

JonathanMatthey:
## Summary

Split Layout now supports a `displayRatio` of `1:1`, `1:2`, or `2:1` (content:media) so editors can match the Figma column proportions.

- Strapi enum + admin layout (Image / Video URL, then position | ratio | alt or URL at 3|3|6)
- Astro uses CSS grid `fr` tracks with the existing 120px (`gap-6xl`) gutter
- MDX sync round-trips the prop (omitted when `1:1`)

## Related Issue

Fixes INTORG-937

## Manual Test

1. In Strapi, open a Split Layout block and confirm Display ratio options: `1:1`, `1:2`, `2:1`
2. Image layout: Image on its own row; Image position | Display ratio | Image alt at 3|3|6
3. Video layout: Video position | Display ratio | Video URL at 3|3|6
4. On the site, check each ratio with media left and right — column widths should match content:media, with a 120px gap at tablet+

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


Infi-Knight:
Rendered this locally across all 3 ratios x 2 positions. The tracks compute correctly, the 120px gutter holds, and mobile still stacks, so the ratio logic is solid.

One thing I want to confirm`1:1` isn't just the neutral default here, it changes existing blocks. The old layout capped media at ~45% / max 22rem (~352px) with content on `flex-1`. The new `1:1` is `grid-cols-2`, a true 50/50 with no cap, so at 1200px the media column jumps from ~352px to 508px. The only page using SplitLayout today is `grant-pages/education/on-campus.mdx`, so blast radius is one page, but its image will get noticeably bigger. If that matches Figma's 50/50, all good. 
Two minor notes, neither blocking:

- The admin row in `configureLayouts` sums to 18 (`imagePosition 3 + displayRatio 3 + imageAlt 6 + videoUrl 6`). Since alt and URL aren't conditionally hidden, both render and the row wraps rather than sitting at 3|3|6. Cosmetic, and the old row overflowed too, just flagging so it's not a surprise.
- Worth a one-line comment on the `SplitLayout.astro` fallback noting that silently coercing an unknown ratio to `1:1` is intentional, since the handler and serializer both hard-fail on the same input.

Infi-Knight (approved):
Please see the previous comment about the ratio, if it's all good let's merge this

JonathanMatthey:
fair point - will default to 2:1 - feels more normal

PR #421, opened 2026-07-23
fix(lint): protect CodeBlock content in MDX prettier lint/format [INTORG-950]

Infi-Knight:
## Summary

Code blocks written in Strapi failed the repo build lint, and a `prettier --write` could quietly corrupt them.

The reason is that the two sides format `CodeBlock` differently. On export the CMS shields the `code={` … `}` attribute from Prettier (`formatMdx` in `cms/src/utils/mdx.ts` swaps it for a placeholder, formats, then puts the code back verbatim). The repo's `prettier --check` / `--write` had no such shield, so it wanted to rewrite the attribute (that is the Strapi vs Astro mismatch) and, when the code contained a blank line, it escaped `*` to `\*` and dropped indentation.

This routes `.mdx` through `scripts/format-mdx.mjs`, which does the same protect, format, restore that `formatMdx` already does (I mirrored its regex and placeholder). `package.json` now runs plain Prettier on the non-mdx globs and the wrapper for `.mdx`, in both `lint` and `format`. `formatMdx` is unchanged.

Result: the repo lint produces the same output as the CMS, so Strapi code blocks pass, and code with blank lines round-trips untouched.

## Related Issue

Fixes INTORG-950

## Manual Test

Verified the whole loop, not just the lint step:

1. Added a `CodeBlock` with a blank line (and a `// spacing * stays` comment) to a post through the local Strapi admin API. Its `afterUpdate` lifecycle exported the MDX into the repo via `formatMdx`, with the blank line and `*` intact and no corruption.
2. `node scripts/format-mdx.mjs --check` passes on that real export; raw `prettier --check` on the same file fails (the pre-fix bug).
3. `pnpm build` is green and the block renders with the blank line as its own numbered line.
4. Confirmed a byte-for-byte match between the wrapper's output and a live `formatMdx` run on the same input.

To reproduce quickly: add `<CodeBlock language="ts" code={` … `} />` with a blank line to any `.mdx`, then run `pnpm lint` (passes) and `pnpm format` (leaves the code untouched).

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

Follow-ups: once this lands, the temporary blank-line stripping in the INTORG-943 blog migration can be reverted so authors may use blank lines again;


Anca2022 (approved):
Great job! :D

PR #425, opened 2026-07-24
fix(blog): fall back to EN for untranslated related articles [INTORG-964]

Infi-Knight:
## Summary

The "Other relevant articles" section did not display on former-developers blog posts (in Spanish). Related-article slugs were resolved strictly in the current locale, so an ES page whose related articles are EN-only (typical for legacy developer posts, which mostly aren't translated) dropped every slug and hid the whole section. Foundation posts' related articles have ES translations, so they still rendered, which is why the two post types looked different.

The resolver now prefers the current locale and falls back to the canonical EN post when no translation exists, matching the site's existing EN-fallback rule (`getLocalizedPaths`). The fallback card links to the EN post, which is the same content an ES visitor would see for that post anyway.

Refactored the logic out of the `.astro` frontmatter into a pure, unit-tested `resolveRelatedPosts` helper.

## Related Issue

Fixes INTORG-964

## Manual Test

1. Set (or sync) one or more `relatedArticles` on a legacy post whose targets are EN-only (e.g. `interledger-universe`).
2. Visit `/es/blog/<slug>` for that post.
3. Before: no "Other relevant articles" section. After: the section renders, showing the EN related posts (linking to their EN routes).
4. Confirm EN routes and ES posts whose related articles have ES translations are unaffected.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/src/components/pages/FoundationBlogPostPage.astro
+++ b/src/components/pages/FoundationBlogPostPage.astro
@@ -1,7 +1,11 @@
 ---
-import { useTranslations, type Locale } from '@/utils'
+import {
+  useTranslations,
+  resolveRelatedPosts,
+  defaultLocale,
+  type Locale
+} from '@/utils'
 import { getCollection, render } from 'astro:content'
-import type { CollectionEntry } from 'astro:content'
 import CommunityLinksFoundation from '@/components/blog/CommunityLinksFoundation.astro'
 import ArticleBio from '@/components/blog/ArticleBio.astro'
 import RelatedArticles from '@/components/blog/RelatedArticles.astro'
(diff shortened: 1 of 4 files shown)


adrianboros (approved):
Looks good!

Only caveat is that - when I tested  - added the en related link to the es page, then created the related es page itself. The parent blog page (for es of course ) didn't pick up the translated related post page until I updated the parent blog page in strapi. Astro cache issue maybe?

Infi-Knight:
@adrianboros thanks! That's expected dev-server behavior rather than a resolver bug. `resolveRelatedPosts` runs at build time, and since the site is `output: 'static'`, every Netlify deploy is a fresh build that picks up new ES translations regardless of whether the parent was touched.

What you hit is Astro's incremental rebuild in dev: adding the ES related post didn't change the parent page's own MDX, so the dev server didn't re-render the parent until your Strapi re-save regenerated its file. On a production build it resolves correctly without needing to touch the parent.

Also pushed a small follow-up (e86b1f61) fixing the Copilot-flagged JSDoc/param mismatch on the helper.

PR #430, opened 2026-07-27
fix(blog): fit tables to the blog width and wrap on tablet+ [INTORG-968]

Infi-Knight:
## Summary

Wide blog tables (e.g. the "Glossary of Terms" table on `/blog/ilf-and-hsms/`) used `width: max-content` and scrolled horizontally at every breakpoint. Per design feedback from Sarah, a table like that should fit the blog content column and wrap its cell text on desktop/tablet, and only fall back to horizontal scroll on mobile (where fitting every column would crush them).

Change: from the `tablet` breakpoint the table is set to `width: 100%`, so auto table layout distributes the columns to fit the blog width and long cells wrap. Mobile keeps `width: max-content` inside `.table-scroll` for the scroll fallback.

Note on the "max column width ~3/4" idea from Sarah: a percentage `max-width` on a `<td>` is not honored by browsers under auto table layout. `width: 100%` already keeps the whole table within the blog column, and auto layout distributes the columns sensibly on its own, so no per-column cap is needed. A hard cap would require `table-layout: fixed` with explicit column widths, which does not fit arbitrary CMS-authored tables. Happy to revisit if a stricter split is wanted.

## Related Issue

Fixes INTORG-968

## Manual Test

Open `/blog/ilf-and-hsms/` and scroll to "Glossary of Terms":
1. Desktop and tablet: the table fits within the blog column and the Description cells wrap to multiple lines. No horizontal scrollbar.
2. Mobile (narrow width): the table keeps its natural width and scrolls horizontally inside its container; the page itself does not overflow.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/src/styles/components/prose/blog.css
+++ b/src/styles/components/prose/blog.css
@@ -37,7 +37,7 @@
       border-inline: 2px solid var(--color-raspberry-100);
       padding-inline: var(--spacing-md);
       margin-block: var(--spacing-2xl);
-      
+
       @variant tablet {
         padding-inline: var(--spacing-lg);
       }
@@ -78,6 +78,11 @@
       outline-offset: 2px;
     }
 
+    /* Mobile keeps the natural (max-content) width so genuinely wide tables can
+       scroll inside .table-scroll — cramming every column into ~340px is worse
+       than a horizontal scroll. From tablet up there's room to fit the table to
+       the blog column, which makes long cells wrap instead of the table
+       scrolling (auto layout then distributes the columns to fit). */
     & table {
       width: max-content;
       min-width: 100%;
@@ -86,6 +91,10 @@
       border-radius: var(--radius-2xl);
       overflow: hidden;
       background-color: var(--color-neutral-0);
+
+      @variant tablet {
+        width: 100%;
+      }
     }
 
     & th,
@@ -170,7 +179,7 @@
     .video-embed {
       border-radius: var(--radius-3xl);
       overflow: hidden;
-      margin-block-end: var(--spacing-2xl)
+      margin-block-end: var(--spacing-2xl);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


src/styles/components/prose/blog.css
```diff
@@ -86,6 +91,10 @@
       border-radius: var(--radius-2xl);
       overflow: hidden;
       background-color: var(--color-neutral-0);
+
```
JoblersTune, line 91: What makes the tablet different that we want width 100 only here?

src/styles/components/prose/blog.css
```diff
@@ -78,6 +78,11 @@
       outline-offset: 2px;
     }
 
+    /* Mobile keeps the natural (max-content) width so genuinely wide tables can
```
JoblersTune: This feels like a very long comment to include in our styling.

JoblersTune (approved):

JoblersTune:
If we have to set it to 100% then we must just keep an eye out because we're losing flexibility. But it does look better this way in this case. I hate managing tables on a CMS 🙈

src/styles/components/prose/blog.css
Infi-Knight: On mobile the column is only ~340px, so forcing the table to fit would squeeze every column into a few characters and stack the text vertically, which reads worse than a horizontal scroll inside `.table-scroll`. From tablet up there's enough width to fit the table to the column and let long cells wrap normally, so we switch to `width: 100%` there. Condensed the comment to say this in two lines (ead38d06).

src/styles/components/prose/blog.css
Infi-Knight: Agreed, trimmed it from five lines to two in ead38d06.

PR #436, opened 2026-07-29
feat(podcast): strapi template and component, mdx page file and astro update

adrianboros:
## Summary
This PR migrates the site’s Podcasts landing page from hardcoded Astro/data to a Strapi-driven MDX content collection, adding the necessary Strapi content type + MDX sync plumbing and updating routing to serve the page at /podcast.

Changes:

- Add a new podcast-pages content collection with frontmatter schema + tests, plus an initial src/content/podcast-pages/podcast.mdx entry.
- Replace the initial `/podcasts` page + static src/data/podcasts.ts with a new routed `/podcast` PodcastPage and supporting shared components, same route as live site.
- Add the Strapi podcast-page content type, MDX generator, sync-mdx transformer support, and populate config.

## Related Issue
Fixes INTORG-692

## Manual Test
- run sync command
- open `/podcast `page

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


src/config/foundation-navigation.json
```diff
@@ -170,7 +170,7 @@
         },
         {
           "label": "Podcasts",
-          "href": "/podcasts"
+          "href": "/podcast"
         },
```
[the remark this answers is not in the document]
adrianboros: that's ok, live site currently points to /podcast

cms/scripts/sync-mdx/mdxTransformer.ts
```diff
@@ -924,20 +988,9 @@ export async function buildBlogPayload(
...
-        featureImageMobile,
-        nullOrValue(parsed.featureImageMobileAlt),
-        updatedAltIds,
-        mdx.pathSlug,
-        dryRun
-      )
-    }
+    // featureImage/thumbnailImage alt text lives on featureMedia/thumbnailMedia
+    // (shared.localized-media); featureImageMobile is a plain media field that
+    // shares the desktop image's alt text, so it has no alt of its own to patch.
```
Anca2022: Not blocking — this looks like cleanup carried over from the PR that introduced shared.localized-media, not new work here. 
But I want to flag a concern: dropping featureImageMobile's own alt and falling back to the desktop image's alt text assumes the mobile image is just a cropped/resized variant of the same photo. If editors ever use a genuinely different image for mobile, the inherited alt text won't describe what's actually shown.

cms/src/utils/podcastPageMdx.ts
```diff
@@ -0,0 +1,150 @@
...
+  MATTER_STRINGIFY_OPTIONS,
+  heroFrontmatter,
+  htmlToMarkdown,
+  type HeroCta
+} from './mdx'
+
+/** CKEditor (basicMarkdownPreset) fields are usually already markdown, but
+ * defensively convert if Strapi ever hands back HTML (matches
+ * generateFaqMdx's identical pattern). */
+function ckeditorFieldToMarkdown(value: string): string {
```
Anca2022: nit: maybe it would be worth moving this function into utils, as this exact function is being declared as well in `cms/src/utils/faqMdx.ts:42`

cms/src/utils/podcastPageMdx.ts
```diff
@@ -0,0 +1,150 @@
...
+export interface PodcastPageHero {
+  title?: string
+  description?: string
+  media?: {
+    image?: MediaFile | null
+    alternativeText?: string | null
+  } | null
+  backgroundImageMobile?: MediaFile | null
+  hero_call_to_action?: HeroCta | null
+}
```
Anca2022: podcastPageMdx.ts:30-39 `PodcastPageHero` duplicates `HeroData` (mdx.ts:201) and `PageData.hero` (pageLifecycle.ts:39) - same Strapi hero component, declared three times, each slightly differently typed. 
That mismatch is why generatePodcastPageMdx:135 needs an as cast to call `heroFrontmatter()`. 
Hoist one shared Hero type into src/types/ and drop the cast.

src/schemas/content.ts
```diff
@@ -281,5 +281,66 @@ export const reportFrontmatterSchema = z.object({
...
+const podcastCtaStripSchema = z.object({
+  heading: z.string(),
+  description: z.string(),
+  buttonText: z.string(),
+  buttonLink: z.string(),
+  color: z.enum(['purple', 'green']).default('purple'),
+  secondaryButtonText: z.string().optional(),
+  secondaryButtonLink: z.string().optional()
+})
+
```
Anca2022, line 297: `podcastCtaStripSchema` duplicates `grantCtaStripSchema` defined above - both back the same Strapi component. Worth merging into a single `ctaStripSchema`.

Same story for `podcastTitleCardSchema`/`podcastTitleCardsSchema` — same shape already hand-typed as `StrapiTitleCard` in `TitleCardGrid.astro`.

Suggest holding off and doing one refactor PR for all of these once every page/component is in, rather than chasing it piecemeal now.

Created an issue for this: INTORG-1017

src/schemas/content.ts
```diff
@@ -281,5 +281,66 @@ export const reportFrontmatterSchema = z.object({
...
+// detectVideoProvider (src/utils/main/video.ts) for how the player branches.
+const podcastItemSchema = z.object({
+  title: z.string().min(1, 'title is required'),
+  description: z.string().min(1, 'description is required'),
+  url: z.string().min(1, 'url is required'),
+  series: z.enum(['Interledger Salon', 'Future Money', 'Off the Ledger'])
+})
+
+// The /podcast landing page. Only one entry is expected to exist; pathSlug is
+// set to "podcast" so it resolves at the site root, no dynamic zone involved.
```
Anca2022: This comment is confusing - it overstates what's actually enforced. pathSlug isn't set to "podcast" by code;  that value comes from the CMS and isn't validated/constrained here. 
And podcast-page is a collectionType, not a singleType, so multiple entries are technically permitted;

src/content/podcast-pages/podcast.mdx
Anca2022, line 1: Please update this page with the content of the final copy (link shared on the ticket)

cms/src/index.ts
Anca2022, line 1: `podcast-page` has no `registerDocumentValidation` call in index.ts, leaving the content validation to Strapi;
we should enforce required-field validation on save, as we do for the other page types

src/components/shared/PodcastList.astro
```diff
@@ -0,0 +1,29 @@
...
+// Authored in chronological order; the newest episode (last in the list) shows first.
+const orderedPodcasts = [...podcasts].reverse()
+---
+
+<ul class="m-0 flex list-none flex-col gap-3xl p-0" role="list">
+  {
+    orderedPodcasts.map((podcast) => (
+      <li>
+        <PodcastItem podcast={podcast} theme={theme} />
+      </li>
```
Anca2022: Please add this requirement from the ticket: 
`Edit: Podcasts require a dividing line between elements as seen on Framer`

Image

src/components/pages/PodcastPage.astro
```diff
@@ -0,0 +1,121 @@
...
+  podcasts,
+  ctaStrip
+} = mdxPage.data
+
+const t = useTranslations(contentLocale)
+
+const breadcrumbs = [
+  { name: t('nav.home'), href: localizeRoute('', contentLocale) },
+  { name: heroTitle ?? title, href: localizeRoute(slug, contentLocale) }
+]
```
Anca2022, line 55:
```
const { routeLocale } = Astro.locals
const t = useTranslations(routeLocale)

const breadcrumbs = [
  { name: t('nav.home'), href: localizeRoute('', routeLocale) },
  { name: heroTitle ?? title, href: localizeRoute(slug, routeLocale) }
]
```

We need to keep navigation in ES, even when we have no ES content; using `contentLocale` here would flip the UI chrome to English whenever content falls back, which is wrong. Please use `routeLocale` instead

src/components/pages/PodcastPage.astro
```diff
@@ -0,0 +1,121 @@
...
+      image={heroImage}
+      imageAlt={heroImageAlt}
+      imageMobile={heroImageMobile}
+      cta={heroCtas?.[0]}
+      breadcrumbs={breadcrumbs}
+    />
+
+    <section class="py-4xl">
+      <div
+        class="mx-auto max-w-content px-4 min-w-90 tablet:px-xl desktop:max-w-content"
```
Anca2022: Please use design token instead of `px-4` value

src/components/shared/PodcastItem.astro
```diff
@@ -0,0 +1,64 @@
...
+---
+
+<article
+  class={twMerge(
+    'flex w-full flex-col gap-xl desktop:flex-row desktop:items-center desktop:gap-2xl'
+  )}
+>
+  <p
+    class={twMerge(
+      'text-base leading-normal mb-0 tablet:text-lg desktop:w-1/2 desktop:text-xl desktop:leading-[30px]',
```
Anca2022: `text-base` / `text-lg` / `text-xl` aren't part of our design system - please swap these for the matching text design tokens (e.g. `h5` family) to match Figma. Those tokens already bundle line-height, so `leading-normal` / l`eading-[30px]` should be dropped rather than set manually.

Anca2022 (commented):
Looks good, Adi! :D 

Left some comments

cms/scripts/sync-mdx/mdxTransformer.ts
adrianboros: Yes, so the idea is that editor should not use genuinely different image for mobile, since the page content is the same, they may want an adapted version of the image for mobile, not a different image content, because if we go on this route with different content then they may want different text content also, so to keep thinks in check one image with alt text, option variants for tablet/mobile. Thoughts @JoblersTune ?

JoblersTune: While it's unlikely that a different image would be used I do think it's possible. Text should remain the same, but let's support alt text per image variant.

JonathanMatthey (approved):
earlier blockers look fixed; LGTM

PR #413, opened 2026-07-22
feat(grants): restructure grant overview pages under /grants [INTORG-849]

JonathanMatthey:
test:  http://localhost:1103/grant/our-grantmaking 

## Summary

- Add a `followUpContent` frontmatter field (synced through the CMS) so grant overview pages can render free-form contact/follow-up markdown below the CTA strip
- Add bottom spacing to stacked `NumberTiles` / `Paragraph` / `TitleCardGrid` blocks
- Minor cleanup: clarify a `sync-images` log message

## Related Issue

Fixes INTORG-849

## Manual Test

- [ ] Open a grant overview page (e.g. `/grant/digital-finance`) and confirm layout/breadcrumbs look correct
- [ ] Confirm the follow-up content section renders below the CTA strip when `followUpContent` is set
- [ ] Confirm stacked number tiles / paragraphs / title card grids have consistent bottom spacing

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] PR title follows Conventional Commits (e.g. `feat: ...`, `fix: ...`)
- [x] Linked issue included
- [x] Scope is focused (target ~10-20 files when possible)
- [ ] Screenshots for UI changes (if applicable)

PR #439, opened 2026-07-29
fix(stats): support a currency prefix on stat tiles [INTORG-993]

Infi-Knight:
## Summary

Grant amounts on the homepage and grant overview pages rendered as a bare number ("21M+"), which means nothing without a currency.

Adds an optional `prefix` to stat tiles, alongside the `suffix` that already exists. Deliberately not a hardcoded `$` or an "is monetary" boolean: it mirrors a field editors already understand, and covers `€`/`£` without another schema change.

The prefix rides *inside* the animating span (`data-prefix`, prepended in the count-up formatter) rather than sitting in a sibling element. Two reasons worth calling out for reviewers:

- A sibling flex child would inherit `gap-sm` and render `$ 21`. Inside the span, the symbol stays glued to the digits.
- The client script swaps the display to `0` before the observer fires and then formats each frame. Without the prefix in that path, a monetary stat would flash a bare `0` and only gain its `$` once the counter settled.

`prefix` is carried through both directions of the Strapi round trip (`number-tiles.serializer.ts` and `numberTilesHandler.ts`), dropping blank values on each side. The affix validation in the parser is now shared between prefix and suffix instead of being suffix-specific.

Applied to the two places from the issue: the homepage grants stat (`$21M+`) and the grant overview tile. Since staging restructured grant overview pages (INTORG-849), that tile now lives in `our-grantmaking.mdx` at `/grant/our-grantmaking`; staging is merged in here and the prefix follows the surviving file. The homepage aria label already read "21 million dollars in grants" in both EN and ES, so no i18n change was needed.

I kept the `+` in the suffix (`$21M+`) since every sibling tile uses it and the issue asked only for the currency symbol. The copy doc line is `$[amount]M in Grants` without it, so that is easy to change if design wants it dropped.

`cms/types/generated/components.d.ts` is checked in, and the change here is the real output of `strapi ts:generate-types`, not a hand edit.

## Related Issue

Fixes INTORG-993

## Manual Test

1. `pnpm build`
2. Open `/` and scroll to "Real-world Progress". The first tile should read `$21M+ / In Grants`, with the `$` tight against the digits and the other three tiles unchanged.
3. Watch the count-up: the `$` should be present the whole way, including the initial frame, and there should be no bare `0`.
4. Open `/grant/our-grantmaking/`. The first tile should read `$21M+ / In Grants`.
5. In Strapi, edit a Number Tiles block: the tile row should read Prefix | Number | Suffix, with help text on the prefix field.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


adrianboros (approved):
👍  
just a note here, `/grant/digital-finance/` has been removed from the content mdx file

Infi-Knight:
Good catch, thanks. That page went away in #413 (INTORG-849) while this branch was open. Staging is merged in as de5e99aa and the merge carried the prefix into the surviving file, `src/content/grant-overview-pages/our-grantmaking.mdx`, which also replaced my `1234` placeholder with the real `21`.

Verified on the preview built from that merge:

- https://deploy-preview-439--interledger-org-v5.netlify.app/grant/our-grantmaking/ renders `$21M+ / In Grants`
- https://deploy-preview-439--interledger-org-v5.netlify.app/ renders the same on the homepage stat

I have updated the Manual Test section to point at `/grant/our-grantmaking/` instead of the removed `/grant/digital-finance/`.

PR #447, opened 2026-07-30
refactor(cms): drop hackathon-page custom generateMDX override [INTORG-987]

JonathanMatthey:
## Summary
Remove hackathon-page's custom `generateMDX` — the shared default now handles pages with no hero and a bare description field.

## Related Issue
Fixes INTORG-987

## Manual Test
- [ ] Save a hackathon page in Strapi and confirm MDX export still looks correct

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/cms/src/api/hackathon-page/content-types/hackathon-page/lifecycles.ts
+++ b/cms/src/api/hackathon-page/content-types/hackathon-page/lifecycles.ts
@@ -1,52 +1,10 @@
-import matter from 'gray-matter'
 import { createPageLifecycle, PATHS } from '../../../utils'
-import type { PageData } from '../../../../utils/pageLifecycle'
-import {
-  defaultLang,
-  MATTER_STRINGIFY_OPTIONS,
-  serializeContent
-} from '../../../../utils'
 import { HACKATHON_PAGE_CONTENT_POPULATE } from '../../../../utils/contentPopulate'
-
-/**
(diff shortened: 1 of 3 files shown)


Infi-Knight:
Not something this PR introduces, but it sits right next to what you touched, so flagging it here.

`heroManagedKeys` in `cms/src/utils/pageLifecycle.ts:198` lists five keys, but `heroFrontmatter` can also emit `heroImageMobile` (`cms/src/utils/mdx.ts:229`). That key never gets deleted from the preserved fields, so if someone removes the mobile background image in Strapi, the stale `heroImageMobile` value stays in the MDX frontmatter forever.

The reason I'm raising it on this PR: your new test is called "clears stale hero fields preserved from MDX when the page has no hero", and it would fail today if you added `heroImageMobile: '/old-mobile.jpg'` to the preserved fields it passes in. As written it reads like it covers the whole set when it covers five of six.

One line:

```ts
const heroManagedKeys = [
  'heroTitle',
  'heroDescription',
  'heroImage',
  'heroImageAlt',
  'heroImageMobile',
  'heroCtas'
] as const
```

Fold it in here or split it out, either is fine by me. It doesn't block the hackathon change, which produces identical output for these pages as far as I can tell (no hackathon MDX currently has any `hero*` key, and the frontmatter key order matches the old override exactly).

Infi-Knight (approved):

PR #451, opened 2026-07-31
feat(scroll): roll out global smooth scroll + add scroll resistance

Anca2022:
## Summary
- adds wheelMultiplier: 0.8 in Lenis settings, for **scroll resistance** 
- anchor-jump handling moves from a custom JS click handler to native `scroll-behavior: smooth` + scroll-margin-top, which now respects prefers-reduced-motion automatically.
- mounts Lenis smooth scroll site-wide via BaseLayout instead of just the homepage
- fixes table scroll-chaining prevention (data-lenis-prevent) to only block horizontal gestures, so vertical wheel/touch scrolling over a table still hands off to Lenis instead of getting stuck.

## Related Issue
Fixes INTORG-978

## Manual Test

 - [ ] Homepage: confirm the AnimatedNetwork section still keeps its slowed-down scroll animation at the end
 - [ ] All pages: scroll should feel "buttery" — scroll resistance applied consistently, not just on homepage
- [ ] overscroll-behavior: none + scroll-behavior: verify the scroll is smooth
  - [ ] /grant/faq — in-page link jump
  - [ ] /blog/web-monetization-open-payments-part-3-sending-money — jump to footnotes
  - [ ] general tab-key scroll behavior 
- [ ] prefers-reduced-motion: reduce — smooth scroll and anchor jumps should fall back to instant/no animation

⚠️  anchor jumps correctly account for the fixed header offset on localhost, but not on deployed environments. Not a quick fix - tracked separately in [INTOG-1005].

## Checks

- [x] `pnpm run format`
- [ ] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/src/components/pages/ContactPage.astro
+++ b/src/components/pages/ContactPage.astro
@@ -118,6 +118,7 @@ const t = useTranslations(routeLocale)
               id="recaptcha-error"
               class="hidden text-red-600 text-sm"
               role="alert"
+              tabindex="-1"
             >
               {t('contact.error.captcha')}
             </p>
@@ -128,6 +129,7 @@ const t = useTranslations(routeLocale)
               id="submit-error"
               class="hidden text-red-600 text-sm"
               role="alert"
+              tabindex="-1"
             >
               {t('contact.error.submit')}
             </p>
@@ -176,6 +178,7 @@ const t = useTranslations(routeLocale)
       )?.value?.trim()
       if (!recaptchaDone) {
         recaptchaError?.classList.remove('hidden')
+        recaptchaError?.focus()
         recaptchaError?.scrollIntoView({ behavior: 'smooth', block: 'center' })
         return
       }
@@ -194,6 +197,7 @@ const t = useTranslations(routeLocale)
       } catch (err) {
         console.error(err)
         submitError?.classList.remove('hidden')
+        submitError?.focus()
         submitError?.scrollIntoView({ behavior: 'smooth', block: 'center' })
       }
     })
--- a/src/components/pages/HomePage.astro
+++ b/src/components/pages/HomePage.astro
@@ -8,7 +8,6 @@ import StatsSection, {
 import TextMediaSection from '@/components/shared/TextMediaSection.astro'
 import LinkButton from '@/components/shared/LinkButton.astro'
 import OptimizedImage from '@/components/shared/OptimizedImage.astro'
-import SmoothScroll from '@/components/shared/SmoothScroll.astro'
 import { useTranslations, type Locale } from '@/utils'
 import PillarsSection from '@/components/shared/PillarsSection.astro'
 import LogoCarousel from '@/components/blocks/LogoCarousel.astro'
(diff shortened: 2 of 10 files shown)


Infi-Knight (commented):
Four things I'd change:

**1. `reset.css:134` is dead code**

`overscroll-behavior: auto` is already the initial value, and nothing in the repo sets `contain` on that attribute (lenis's own CSS does, but we never import it). If it ever did win the cascade it would clobber `overscroll-behavior-x: contain` on `.table-scroll` in `prose/blog.css:70`, which is what stops macOS back-swipe at the table's edge. It only loses today because `components` beats `base`, which is a thin reason for a rule to be harmless. I'd delete it and put the explanation as a comment on the `[data-lenis-prevent]` rule above.

**2. `:target` is narrower than the problem**

It only matches while the URL fragment points at the element, so it misses tab-key focus scrolls (which is on your own test list) and `scrollIntoView` calls. `ContactPage.astro:179,197` scrolls form errors into view and will land them under the header. Something like this covers more ground:

```css
:where(:target, :is(h1, h2, h3, h4, h5, h6)[id], [id]:focus-visible) {
  scroll-margin-top: var(--site-header-height);
}
```

**3. `@media (min-width: 1200px)` duplicates the `desktop` screen**

The value is right, but reset.css already uses `@variant desktop { ... }` in the `hr` rule further down. Worth matching so the two can't drift.

**4. `overscroll-behavior: none` on `html` isn't in the summary**

It kills pull-to-refresh on Android Chrome and back/forward swipe on iOS Safari, site-wide. That may be exactly what you want alongside Lenis, but it deserves a line in the description and a mobile test step. `overscroll-behavior-x: none` gets the swipe protection without touching the vertical axis.

Smaller stuff:

- `touchMultiplier` is still 1, so the resistance is wheel-only. Fine if deliberate, just not what "all pages should feel buttery" implies on mobile.
- `0.8` inline next to a named `DURATION_SECONDS` reads a bit odd. A `WHEEL_MULTIPLIER` const would match the file.
- Lenis now runs on every BaseLayout page, so worth eyeballing the sticky headers in `RoadmapBoard.astro:110` and `FaqSectionsNav.astro:21`. Starlight docs don't use BaseLayout and keep native scrolling, assuming that's intentional.

On INTORG-1005: my guess is that native fragment scrolling fires at parse time, before late-loading images and fonts settle, so layout shifts after the jump. The old JS handler ran after DOMContentLoaded and masked it. If that's the cause, the fix is reserving space for images rather than tuning scroll-margin-top. Throttling the network locally should reproduce it.

Infi-Knight (approved):

Anca2022:
Thank you for your thorough review, Ravi! :D 
Adressed 
1. reset.css:134 is dead code
2. :target is narrower than the problem
3. @media (min-width: 1200px) duplicates the desktop screen
4. overscroll-behavior: none on html isn't in the summary - removed `overscroll-behaviour` altogether from html, left Lenis handle it

Smaller things:
- `touchMultiplier` is still 1 - `touchMultiplier` is a no-op without `syncTouch: true`. Enabling `syncTouch` replaces native touch momentum with Lenis's own inertia curve, and Lenis's README flags it as **unstable on iOS < 16**. Not worth that risk for this PR, so touch stays native pn mobile.
- 0.8 inline next to DURATION_SECONDS reads a bit odd - moved value into const
- worth eyeballing the sticky headers in RoadmapBoard.astro:110 and FaqSectionsNav.astro:21 - ✅

PR #452, opened 2026-07-31
fix(umami): improve umani tracking logic

adrianboros:
## Summary
Root cause: Umami's tracking script blocks navigation on every same-tab `<a>` click carrying `data-umami-event` — it calls `preventDefault()`, waits for a `fetch() `to `ilf-site-analytics.netlify.app/api/send` (~300-500ms+, more on the first click per session due to CORS preflight), and only then navigates. Every nav link had that attribute, so ordinary in-site navigation paid the same tax meant for outbound links.

Fix: Nav/menu links now carry `data-track-event*` instead of `data-umami-event*`. A new click handler in `src/scripts/header-nav.ts` fires the identical event via `window.umami.track()` — which just does a `fetch(..., {keepalive:true})` with no `preventDefault()` — so navigation is never blocked, and no analytics data is lost (verified against ADR-005/006, which explicitly rule out dropping tracking).

Changes:

- `src/utils/main/umami.ts` — added `buildDeferredUmamiAttrs()`, same event/properties, different attribute prefix
`src/scripts/header-nav.ts` — added `initNavClickTracking()`
- 7 nav components switched from `buildUmamiAttrs` to `buildDeferredUmamiAttrs`
- Tests added in `umami.test.ts`; rich-text/MDX links and `PdfEmbed's` download link were deliberately left untouched (out of scope, no delay problem there today)

## Related Issue
Fixes INTORG-851

## Manual Test
1. Locally (`http://localhost:1103/`): open devtools Network tab, click a nav link, confirm navigation happens immediately (no `preventDefault`/await gating it), and confirm a `sendBeacon` request to the collection endpoint is still visible in the Network tab (beacons show as `fetch`/`ping` type requests depending on browser) firing alongside the immediate navigation.
2. Deploy to a preview/staging build and repeat the same check with network throttling (e.g. "Slow 3G") to simulate the worst case that was originally reported — navigation should no longer wait on the throttled connection.
3. Confirm the nav click event still appears in the Umami dashboard with the **same** name/properties as before this change (`nav` label, `base_component: menu`, correct `link_text`, `lang`, `current_path`, `destination_path`, `destination_section`) — this change must be purely a delivery-mechanism swap, not a schema or data change per ADR-005/006.
4. Confirm pageviews still show up correctly in the Umami dashboard for the affected internal pages (pageview beacon is a separate, unrelated mechanism and unaffected by this change).
5. Confirm rich-text/MDX inline links and `PdfEmbed`'s download fallback link (left on the old Umami automatic-attribute path, out of scope for this fix) are unaffected — still tracked, still on their existing delivery path.
6. Confirm `target="_blank"` / new-tab links still record click events as before (they were never delayed, so no perf change expected there — just confirm tracking is untouched).

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


JonathanMatthey (approved):
Looks good — deferred nav tracking is the right fix; LinkButton “caller attrs win” covers the Foundation CTA case

PR #454, opened 2026-07-31
fix(blog): remove stale thumbnails and fix Strapi sync clearing

Anca2022:
## Summary
- Add in new default developer blog thumbnail
- Removed 81 outdated blog thumbnail images and their `thumbnailImage`/`thumbnailImageAlt` frontmatter fields from all 114 foundation-blog-posts 
(they had the wrong aspect ratio for the redesigned card layout and will be replaced later.)
- Fixed the MDX→Strapi sync (`sync:mdx`) so a cleared `thumbnailImage` actually clears `thumbnailMedia` in Strapi (was previously left stale). 
`featureMedia` keeps its omit-when-absent behavior since it's a required field in Strapi.
- Restored `thumbnailImage` as the primary source in `getBlogThumbnail`, now safe since the mismatched thumbnails are gone; posts fall back to `featureImage` until new thumbnails are added.

## Related Issue
Fixes INTORG-886
Fixes INTORG-986

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included


cms/scripts/sync-mdx/mdxTransformer.ts
```diff
@@ -981,7 +981,7 @@ export async function buildBlogPayload(
       featured: parsed.featured ?? false,
       ...(featureMedia ? { featureMedia } : {}),
       featureImageMobile,
-      ...(thumbnailMedia ? { thumbnailMedia } : {}),
+      thumbnailMedia,
```
Infi-Knight, line 986: Worth a comment here. `featureMedia` two lines up is conditional and this one isn't, and the reason (featureMedia is `required: true` in the schema so Strapi rejects null, thumbnailMedia is optional so null is what clears it) currently only lives in the test names.

Infi-Knight (commented):
One thing to fix before this merges: `BlogCard.astro:94` still reads alt text from a single field.

```astro
alt={blogData.thumbnailImageAlt ?? ''}
```

`getBlogThumbnail` falls back thumbnailImage -> featureImage -> tech fallback, but the alt doesn't follow. Since this PR strips every `thumbnailImageAlt` from frontmatter, every card on the listing, featured strip and related-articles rail now renders `alt=""`, while 98 posts still carry a non-empty `featureImageAlt` that goes unused. A `getBlogThumbnailAlt` next to `getBlogThumbnail` with the same precedence would keep the two in step once new thumbnails land.

The rest holds up: the frontmatter edits are clean (195 removed lines, all thumbnail fields, zero additions), every post left without a `featureImage` is `legacy: true` so nothing loses its card image, and nothing in the repo still points at the 83 deleted assets.

Infi-Knight (approved):

cms/scripts/sync-mdx/mdxTransformer.ts
Infi-Knight, line 986: The reason this one is nulled while `featureMedia` two lines up is omitted only lives in a test title. Two lines of comment here would stop someone "fixing" the inconsistency in six months: `thumbnailMedia` is optional in the schema so `null` clears it, `featureMedia` is `required: true` so `null` gets rejected.

src/utils/main/blog.ts
```diff
@@ -12,7 +12,7 @@ export const FEATURED_POST_LIMIT = 3
  * the tech and foundation blogs merge (INTORG-691). Swap this path if comms
  * provides a dedicated asset.
  */
-export const TECH_BLOG_FALLBACK_THUMBNAIL = '/img/og-developers.png'
+export const TECH_BLOG_FALLBACK_THUMBNAIL = '/img/tech-thumbnail.svg'
```
Infi-Knight, line 11: The comment above still says "the current Engineering blog image" and "Swap this path if comms provides a dedicated asset". Comms has now provided one, so this needs a refresh.

cms/types/generated/contentTypes.d.ts
```diff
@@ -1078,8 +1078,7 @@ export interface ApiHackathonPageHackathonPage
         'blocks.split-layout',
         'blocks.profile-grid',
         'blocks.title-card-grid',
-        'blocks.carousel',
-        'blocks.footer-notes'
+        'blocks.carousel'
```
Infi-Knight, line 1081: Unrelated to this PR and it looks alarming, so flagging that I chased it and it's fine. `blocks.footer-notes` doesn't exist anywhere at this head: no `footer-notes.json` in `cms/src/components/blocks`, and `hackathon-page/schema.json` doesn't list it. Staging's committed types are the stale ones and this is a correct regeneration (the same drift shows up in my own working tree).

Worth a note in the description so the next reviewer doesn't burn time on it. Separately, someone should regenerate on staging so this stops riding along on unrelated PRs.

cms/scripts/sync-mdx/mdxTransformer.test.ts
```diff
@@ -2839,7 +2839,7 @@ describe('buildBlogPayload', () => {
         updatedAltIds
       )
 
-      expect(payload).not.toHaveProperty('thumbnailMedia')
+      expect((payload as Record<string, unknown>).thumbnailMedia).toBeNull()
```
Infi-Knight, line 2843: The cast steps around the `| Error` narrowing, so if `buildBlogPayload` ever returned an Error this reads `undefined` and fails with a confusing message. An `expect(payload).not.toBeInstanceOf(Error)` first would make the failure legible.

Infi-Knight (commented):
Reviewed at the PR head in a worktree. The cms suite and the Astro blog tests pass locally, CI agrees.

I checked the deletion sweep carefully since 82 files is a lot to eyeball. Nothing in the repo still references a removed asset, the optimized-image manifests aren't git tracked so there's nothing stale to regenerate, and the frontmatter diffs are clean (just the two lines, no collateral). Coverage works out too: 108 posts have a `featureImage`, 35 are `legacy` with none, which is all 143. Nothing falls through to `null` and renders imageless.

The sync fix is right, and I confirmed the asymmetry is deliberate rather than accidental: `featureMedia` is `required: true` in the foundation-blog-post schema and `thumbnailMedia` isn't, so nulling the first would fail validation. Round trip is safe as well, `blogLifecycle` omits `thumbnailImage` when `thumbnailMedia` is null, so a Strapi save won't put the field back.

Two things before merge.

`BlogCard.astro:94` is `alt={blogData.thumbnailImageAlt ?? ''}`. Every post just lost that field, so it evaluates to `''` for all 143 cards while the src is now `featureImage`, which has a perfectly good `featureImageAlt` going unused. Not a regression (staging already paired `featureImage` with `thumbnailImageAlt`) and an empty alt next to a linked title is defensible, but the expression is dead now and nobody can tell whether that's deliberate. Either hardcode `alt=""` with a comment saying the linked title carries the meaning, or have `getBlogThumbnail` return `{ src, alt }` so the alt follows whichever source won. I'd take the first, given the 35 legacy posts all render the same default SVG.

Second, worth a line in the description: once this lands and `sync:mdx` runs, a thumbnail set through the Strapi admin gets wiped on the next sync. That matches "Astro is the source of truth", but whoever adds the replacement thumbnails needs to know they go in MDX frontmatter, not the admin UI.

Minor stuff: `public/img/foundation-blog/testnet-wallet-thumbnail.svg` survived the sweep and is now referenced by nothing, worth deleting with the rest. The new default is 1200x630 against cards at `aspect-2/1` and `aspect-240/140`, so `object-cover` will crop it. Probably fine, but 35 posts share that one image, so give it a look on the deploy preview at both card sizes. The file itself is clean: minified, 17.9 KB, no C2PA metadata.

Commit messages don't follow Conventional Commits and carry a couple of typos ("[rimary srouce", "feeatureMedia"). Moot if you squash on the PR title, which is fine as-is.

PR #444, opened 2026-07-29
feat(astro): add profile component to hackathon templates and demo pages

adrianboros:
## Summary
- add profile grid component to hackathon template dynamic zone
- add profile grid component to demo pages

## Related Issue
Fixes INTORG-912

## Manual Test
run sync command
open any of the demo pages in strapi and astro /demo-foundation-page , /demo-summit-page or /demo-hackathon-page

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] PR title follows Conventional Commits (e.g. `feat: ...`, `fix: ...`)
- [x] Linked issue included
- [x] Scope is focused (target ~10-20 files when possible)
- [ ] Screenshots for UI changes (if applicable)

PR #456, opened 2026-08-03
feat(astro,strapi): add FAQ block to page dynamic zones [INTORG-909]

Infi-Knight:
## Summary

Adds a `blocks.faq` component (optional heading, repeatable question/answer items) to the foundation, summit and hackathon dynamic zones, with both render paths wired and a full MDX round-trip.

A few decisions worth knowing about:

- **Reuses `blocks.faq-item`, not `blocks.faq-section`.** `blocks.faq-section` is nearly what's needed already (required heading + repeatable `blocks.faq-item`), but its heading drives the Grantmaking FAQ page: `FaqSections.astro` derives each section's anchor id from it, and `FaqSectionsNav` renders it as the left-nav label. Relaxing it to optional wouldn't crash that page, there's a `|| 'section'` slug fallback, but it would leave blank nav entries and empty `<h2>`s unless both components were reworked. So `blocks.faq` nests the same `blocks.faq-item` child and leaves `faq-section` untouched.

  Downside worth naming: `blocks.faq` and `blocks.faq-section` now look nearly identical in the admin's component picker. Collapsing them into one component with an optional heading, and teaching the FAQ page to handle the empty case, is the cleaner end state if someone wants to pick it up.
- **Items are JSX children, not a prop.** Follows `TitleCardGrid` rather than `NumberTiles`, because the answer is a CKEditor rich-text field and markdown in a JSON attribute is miserable to author.
- **Every item renders collapsed.** The mock shows the first one open, and both existing FAQ surfaces open index 0. But in the MDX path items arrive through a `<slot>`, so marking "first item" would need either string surgery on rendered HTML or a Strapi preview that diverges from the published page. Collapsed keeps both paths identical, with no JS and no layout shift.

### Also worth a look

`cms/types/generated/contentTypes.d.ts` drops `blocks.footer-notes` from the hackathon zone. Worth a sentence of history, because this is the third time it's been removed:

- #6cc84216 added a footer-notes component
- #b333271f reverted it a day later in favour of the existing GFM footnote handling, deleting every file
- #891a780d regenerated the types to drop it
- #958dd204 (PR #444) put it back into the generated types, presumably regenerated on a machine whose Strapi still had it registered
- Anca removed it again in #bb61acaa, but that's on `am/blog-thumbnail` and hasn't reached staging

So `main` is clean, `staging` still carries it, and this branch removes it again. No schema or component file for it exists anywhere in the repo, so hand-editing it back would mean committing generated output that contradicts the schemas. Flagging it mainly because it'll conflict with Anca's branch if both land.

## Related Issue

Fixes INTORG-909

Also filed while working on this: INTORG-1013, where editing any foundation/summit/hackathon page in Strapi and hitting Publish deletes its `SplitLayout` and `LogoCarousel` blocks from the MDX. Pre-existing, unrelated to this work, and deliberately kept out of this PR to keep it scoped. It's a live content-loss bug on staging, so worth prioritising separately.

## Manual Test

1. `pnpm start`, then visit `/demo-foundation-page` and scroll to "About the Interledger Foundation"
2. Click a question. Only that one opens, the chevron rotates, the answer renders markdown including links
3. `Tab` through the questions. Each `<summary>` takes focus with a visible orchid ring
4. Narrow to mobile. Type scales down, no horizontal overflow
5. `/hackathon/demo-hackathon-page` has a heading-less FAQ. It renders no `<h2>` and the section falls back to a translated `aria-label` (`Preguntas frecuentes` on `/es/hackathon/...`)

Round-trip, if you want to verify it: sync the MDX into a local Strapi, edit the page in the admin, publish, and diff the MDX. The FAQ block comes back byte-identical (verified at 953 B, 434 B and 380 B on the three page types).

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [ ] Scope is focused 


src/components/blocks/DynamicZone.astro
```diff
@@ -54,6 +55,7 @@ const componentMap: Record<string, any> = {
   'blocks.split-layout': SplitLayout,
   'blocks.number-tiles': NumberTiles,
   'blocks.title-card-grid': TitleCardGrid,
+  'blocks.faq': Faq,
```
Anca2022, line 60: `blocks.faq` is mapped here, but preview routes won't render it (`page-preview.astro` and `preview/page-preview.astro`).

Please add a `populate[content][on][...]` query for the dynamic zone inside these files.

src/components/shared/FaqItem.astro
```diff
@@ -0,0 +1,41 @@
...
+    >
+      {question}
+    </span>
+    <Icon
+      name="arrow-down"
+      class="size-8 shrink-0 text-orchid-100 transition-transform duration-200 group-open:rotate-180 motion-reduce:transition-none"
+    />
+  </summary>
+  <div
+    class="mt-md text-h5 tablet:text-h5-md desktop:text-h5-lg text-neutral-75 [&_p]:m-0 [&_p+p]:mt-md [&_:last-child]:mb-0"
```
Anca2022: The `text-h5` and `text-neutral-75` classes applied to the `<div>` won’t be inherited by the `<p>` elements inside it. You need to explicitly scope them, e.g. `[&_p]:text-h5`, because the **p, a, ol**, and **ul** rules defined in `src/styles/components/prose/default.css:13` have higher specificity in the cascade. 
The fact that the text currently renders as `text-h5` is coincidental - it’s matching the corresponding rule in components.

Anca2022 (approved):
- Left 2 comments, small stuff

- Found a bug - not blocking, as long as you either fix it or track it in an issue - that's up to you :)
   Bug: when you add bulleted list inside `answer` the rendering of the page breaks 


Image

the rest looks good :D

src/components/blocks/DynamicZone.astro
Infi-Knight: Fixed. Those routes build their own `on` map instead of reusing `contentPopulate.ts`, so the block never came back from the query at all.

Added to both:

```ts
'populate[content][on][blocks.faq][populate][items]': '*'
```

Checked against a local Strapi and the block shows up in preview now.

Both routes are also missing `split-layout`, `title-card-grid` and `video-embed`, and `page-preview.astro` is missing `pdf-embed` on top of that, so those preview incomplete too. Leaving them out of this PR.

src/components/shared/FaqItem.astro
Infi-Knight: You're right, and it was worse than cosmetic. The answer text was rendering at `neutral-100`:

| | font-size | colour |
| -- | -- | -- |
| wrapper `<div>` | 20px | `rgb(89,89,89)` |
| actual `<p>` | 20px | `rgb(41,41,41)` |
| design | 20px | `rgb(89,89,89)` |

The size matched only because `main p, a, ul, ol` in `prose/default.css:13` uses the same `--text-h5` token, so it would have drifted the moment that token changed. When I checked this originally I measured the wrapper instead of the paragraph, which is how it got past me.

Now scoped per element with `[&_p]`, `[&_li]` and `[&_a]`, with a note on the component so nobody reaches for inheritance here again. `p` and `li` both come out at 20px / `rgb(89,89,89)`.

Infi-Knight:
On the list bug: two separate things were wrong, both fixed.

First was the colour mismatch from the `FaqItem` thread. Prose CSS styles `p` but not `li`, so paragraphs took `neutral-100` from the cascade while list items inherited `neutral-75` from the wrapper. One answer, two colours.

Second is the one that actually looks broken, and I missed it on the first pass. A list with blank lines between items produces `<li><p>text</p></li>`. `prose/default.css:52` sets `list-style-position: inside`, so the marker shares the first line box of the item's content, and a block-level `<p>` pushes it onto its own line. You end up with the bullet or number sitting alone above its text.

| Variant | `<li>` first child | `<li>` height |
| -- | -- | -- |
| tight | `TEXT` | 30px |
| loose, before | `P` | 60px |
| loose, after | `P` (`display: inline`) | 30px |

Every list I tried the first time was tight, which is why I thought it was already sorted. Inlining the item's first paragraph fixes it.

Verified on both render paths, since they're different code: MDX children, and the Strapi `items` prop going through `parseMarkdown` into `set:html`.

One thing I left alone: the markers sit flush left with no indent. That comes from `prose/default.css:51-53` setting `list-style-position: inside` with `padding-inline-start: 0`, and a Paragraph block with the same list looks identical, so it's site-wide rather than this block.

Anca2022 (approved):
looks great! Thank you, Ravi! :D

PR #458, opened 2026-08-03
feat(INTORG-908): support purple CTA strip component - INTORG-908

JonathanMatthey:
## Summary
- Enabled CTA strip rendering for Hackathon pages.
- Removed remaining CTA strip color support from grant sync/content.
- Updated previews/tests for purple-only, primary CTA-only strips.

## Related Issue
Fixes INTORG-908

## Manual Test
Not needed; covered by build and tests.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused


JonathanMatthey:
image

image

cms/scripts/sync-mdx/mdxTransformer.ts
Infi-Knight, line 956: `blocks.cta-strip` is missing from `HACKATHON_PAGE_ALLOWED_COMPONENTS `, this will fail on sync

Infi-Knight: add 'blocks.cta-strip' to the list, and add a buildHackathonPagePayload test covering it, the allowlist has drifted from schema.json silently so a test that asserts the two agree would be worth more than one that hardcodes the array.

cms/src/serializers/blocks/cta-strip.serializer.ts
```diff
@@ -3,51 +3,38 @@ import { htmlToMarkdown } from '../../utils'
...
 
 export function serialize(block: {
-  heading: string
-  description: string
+  heading?: string
+  description?: string
   primaryButtonText: string
   primaryButtonLink: string
   secondaryButtonText?: string
   secondaryButtonLink?: string
```
Infi-Knight, line 19: The param type still declares secondaryButtonText?, secondaryButtonLink?, color?, none of which are read. the schema no longer returns them, so they're just noise now

Infi-Knight: also Validation (`cms/src/utils/contentValidation.ts`) no longer catches an empty strip. With heading and description both optional in `validateCtaStrip` and in the Astro component, a strip with only a button is now valid at every layer, rendering as a bare button on a colored box. Consider requiring at least one of heading/description:

```
if (!heading?.trim() && !description?.trim()) {
  fieldErrors.push({
    message: 'CTA Strip: Provide a heading or a description',
    path: ['ctaStrip', 'heading']
  })
}
```

Infi-Knight (changes requested):

cms/scripts/sync-mdx/mdxTransformer.ts
JonathanMatthey: good catch - done

JonathanMatthey:
@Infi-Knight  review again please

Infi-Knight (approved):

PR #459, opened 2026-08-03
feat(site): add card grid dynamic zone demos (INTORG-885, INTORG-889, INTORG-890)

JonathanMatthey:
## Summary

- Added the unified Card Grid model, sync support, serializers, and Astro rendering.
- Added demo content for foundation, summit, and hackathon pages to preview dynamic-zone components.
- Updated summit and hackathon page content styling, including light default text and card heading levels.

## Related Issue

Fixes INTORG-885

## Manual Test

- Visit `/demo-foundation-page`, `/summit/demo-summit-page`, and `/hackathon/demo-hackathon-page`.
- Confirm the dynamic-zone previews render, including the new card grids.
- Confirm summit and hackathon body text defaults to light text where expected.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

image


cms/src/utils/cardGrid.ts
```diff
@@ -0,0 +1,100 @@
...
+export const CARD_GRID_VARIANT_DEFINITIONS = [
+  {
+    value: 'Info',
+    label: 'Info',
+    cardsField: 'infoCards',
+    fieldLabel: 'Info cards',
+    strapiComponent: 'blocks.info-card',
+    mdxChild: 'InfoCard',
+    helpText: 'Add Info cards for this grid.'
+  },
```
Anca2022: nit: helper text isn't being picked up; probably because each card is a component, not a field.

Might as well remove it

cms/scripts/sync-mdx/cardGridHandler.ts
Anca2022, line 1: 🔥 🔥

cms/src/api/grant-overview-page/content-types/grant-overview-page/schema.json
```diff
@@ -81,7 +81,7 @@
         "blocks.cta-strip",
         "blocks.carousel",
         "blocks.number-tiles",
-        "blocks.title-card-grid"
+        "blocks.card-grid"
```
Anca2022, line 84: For Grant Overview Page the requirement was to have TitleCardGrid inside dynamic zone, not all kinds of cards

See INTORG-849

Please swap back to `title-card-grid` for this template page
Also, please bring back the serializers and handlers for TitleCardGrid - we will need those for Podcast template as well (currently in review)

Anca2022 (commented):
**Part 1 of the review: Strapi side**

⚠️ Please bring back TitleCardsGrid - it's being currently used on the dynamic zone in Grant Pages Overview (Ticket INTORG-849 only requires TitleCards are supported, not all 4 types prezent in CardGrid)

nit: helperText for the 4 Card components is deadcode (cms/src/utils/cardGrid.ts:12)


Other than that, on the cms side this component looks fantastic! Great job! 🎉 🚀

JonathanMatthey:
this is the big one @Infi-Knight  - review again

anca did quite a bit already

src/components/shared/ResourceCard.astro
```diff
@@ -0,0 +1,83 @@
...
+const {
+  heading,
+  buttonUrl,
+  buttonText,
+  buttonExternal = false,
+  buttonDocument = false
+} = Astro.props
+
+const { routeLocale } = Astro.locals
+const href = buttonExternal ? buttonUrl : ensureLeadingSlash(buttonUrl)
```
Infi-Knight: If `buttonDocument=true` and `buttonExternal=false`, this falls into `ensureLeadingSlash(buttonUrl)`, which only checks `startsWith('/')`. An absolute document URL (e.g. `https://partner.org/whitepaper.pdf`) becomes `/https://partner.org/whitepaper.pdf`, a broken link. Validation currently only rejects `external && document` together, not `document` combined with an absolute URL. Same pattern in `NavigationCard.astro:27` and `TitleCard.astro:26`.

cms/src/serializers/blocks/card-grid.serializer.ts
```diff
@@ -0,0 +1,343 @@
...
+    }
+    return block
+  }
+
+  // Active field empty (or variant missing) but cards exist in exactly one
+  // other field — adopt that field's variant instead of wiping the only
+  // cards an editor added into a mismatched section.
+  if (populated.length === 1) {
+    const field = populated[0]!
+    block.variant = FIELD_TO_VARIANT[field]
```
Infi-Knight, line 205: This fallback rewrites `block.variant` before `validateCardGrid`'s cross-field column check runs. Example: `variant: 'Title'`, `columns: 'One'` (invalid, only Navigation may use one column), `titleCards` empty, and a stray populated `navigationCards` (a real state since the admin picker deliberately leaves inactive fields uncleared on mount to dodge a hydration race). This silently flips `variant` to `'Navigation'`, and `validateCardGrid` then validates the rewritten variant and passes. The save that should have failed for what the editor actually selected succeeds as something else instead.

cms/scripts/sync-mdx/jsxExtract.ts
```diff
@@ -704,3 +704,31 @@ export function getChildElements(
...
+ * getMismatchedChildElements(node, 'InfoCard')  // → [titleCardNode]
+ * ```
+ */
+export function getMismatchedChildElements(
+  node: JsxBlockNode,
+  name: string
+): JsxBlockNode[] {
+  const elements: JsxBlockNode[] = []
+  for (const child of node.children) {
+    const element = toJsxElement(child)
```
Infi-Knight: `toJsxElement` (used here and in `getChildElements`) only unwraps a paragraph into a JSX element when it has exactly one child. If a card tag shares a paragraph with any sibling content, e.g. `<ResourceCard/>Oops <ResourceCard/>` on one line, it returns `undefined` for that paragraph and it's dropped from parsing entirely. Both this function and `getChildElements` never see those cards, so the author gets a misleading "requires at least one/two cards" error even though cards were authored.

src/components/blocks/CardGrid.astro
```diff
@@ -0,0 +1,191 @@
...
+      gridColumnsClass
+    ]}
+  >
+    {
+      htmlCards ? (
+        htmlCards.map((card) => {
+          if (variant === 'Title') {
+            if (!hasSecondaryCta(card)) {
+              warnMissingSecondaryCta(card, variant)
+              return null
```
Infi-Knight: Cards missing a valid `secondaryCta` are dropped from the grid here (and at lines 151, 168) with only a server-side `console.warn`, no on-page indication. The comment above says this is "defense-in-depth only" since `validateCardGrid` should catch it at save time, but anything that bypasses that path, content synced from MDX, a script-driven import, older content saved before the check existed, renders a lopsided grid with no signal to editors or visitors that a card is missing.

cms/scripts/sync-mdx/cardGridHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+        component: 'CardGrid',
+        prop: 'columns',
+        line: node.position?.start.line,
+        column: node.position?.start.column
+      })
+    }
+
+    const childName = CARD_GRID_VARIANT_CHILDREN[variantAttr]
+    const cardNodes = getChildElements(node, childName)
+    if (cardNodes.length === 0) {
```
Infi-Knight: This zero-count check runs before the mismatched-component check below. So `<CardGrid variant="Info">` containing only `<TitleCard>` children throws "requires at least one `<InfoCard>` child" instead of "only accepts `<InfoCard>` children, found `<TitleCard>`". The author gets told to add a card when the real fix is swapping the component. Worth swapping the order so the more specific error wins.

cms/src/plugins/card-variant-picker/admin/CardVariantPicker.tsx
```diff
@@ -0,0 +1,327 @@
...
+/** Every card repeatable on the schema — used to show/hide/clear inactive ones. */
+const ALL_CARD_FIELDS: CardGridCardsField[] = CARD_GRID_VARIANTS.map(
+  (variant) => CARD_GRID_VARIANT_FIELDS[variant]
+)
+
+/**
+ * Content-manager URLs look like:
+ * /admin/content-manager/collection-types/api::grant-page.grant-page/...
+ */
+function getContentTypeUidFromLocation(): string | null {
```
Infi-Knight, line 36: This returns `null` whenever the admin URL doesn't match the exact `content-manager/(collection|single)-types/api::...` shape, and `getAllowedCardGridVariants(null)` fails open by returning every variant. Any admin surface that edits this field outside that exact URL shape (a relation/modal editor, a custom view, a future Strapi route change) silently unlocks every variant for content types meant to be restricted, with nothing shown at selection time. Might be safer to fail closed (restrict to the intersection or block editing) when the UID can't be resolved.

src/components/shared/LinkButton.astro
```diff
@@ -96,4 +99,5 @@ const umamiAttrs = buildUmamiAttrs({
   <slot />
   <slot name="icon-right" />
   {external && <span class="sr-only">{t('aria.opens_new_tab')}</span>}
+  {isDocument && <span class="sr-only">{t('aria.downloads_file')}</span>}
```
Infi-Knight, line 134: The `document`/`isDocument` prop only adds a screen-reader-only "(downloads a file)" hint, it never sets the native `download` attribute on the anchor. The Resource card CTAs in the hackathon demo content actually use `buttonDocument={true}` pointing at a PDF, so right now screen reader users hear "downloads a file" but the browser just renders the PDF inline unless the server happens to send `Content-Disposition: attachment`, which this component doesn't control.

Infi-Knight (commented):
Went through the card grid changes end to end (including a run through the demo hackathon page in the browser). Left inline notes on the correctness and accessibility issues I found; the rest were maintainability nits I'll mention separately.

JonathanMatthey:
comments addressed

Infi-Knight (approved):

PR #462, opened 2026-08-05
feat(blocks): add Quote component for pages and blog [INTORG-910]

JonathanMatthey:
## Summary

Adds a new **Quote** block for foundation, summit, and hackathon pages plus foundation blog posts: required quote text with optional author name, image, and link. Also aligns rich-text blockquotes with the same dual-edge design, ships demo MDX pages for visual QA, and keeps those demos off SEO (`noindex` + sitemap filter).

## Related Issue

Fixes INTORG-910

## Manual Test

1. Open foundation / summit / hackathon **demo** pages and confirm both Quote variants (avatar + link above, name-only attribution below).
2. Confirm markdown `>` blockquotes pick up dual raspberry borders and type scale.
3. In Strapi, add a Quote block and check author name, link, and image sit on one admin row.
4. View page source on a demo page: `robots` is `noindex, nofollow`; demos are absent from the sitemap.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused (target ~10-20 files when possible) — slightly over as this wires Astro + Strapi end-to-end
- [ ] Screenshots for UI changes (if applicable) — use deploy preview of demo pages


Infi-Knight:
## Review

**Blocking: Blockquote will double-render quote marks for CMS-synced content.**

`formatBlockquote()` in `cms/src/utils/mdx.ts` (unchanged here) wraps exported quotes in literal curly quotes, so synced MDX looks like:

```
<Blockquote source="...">
  "Money should move like data."
</Blockquote>
```

`Blockquote.astro`'s new `.quote-slot` styling adds its own `::before`/`::after` curly quotes on top of that, regardless of whether the content is `<p>`-wrapped. I verified this in the browser with a fixture matching real sync output: it renders as `""Money should move like data.""`, doubled quotes on both ends.

The demo pages added in this PR don't catch it because they only use the `quote` prop path and markdown `>` blockquotes, never the JSX-children path that real Strapi-synced content actually goes through. `Quote.astro` doesn't have this problem since its serializer doesn't add literal quote characters at export.

Fix: pick one source of truth for the marks, either drop the CSS pseudo-quotes on the slot path or stop `formatBlockquote` from adding them.

**Non-blocking:**

- `astro.config.mjs`'s sitemap filter reimplements the segment-matching logic that's already in `isDemoPathname` (`src/utils/shared/demoPaths.ts`) instead of importing it. The two have already drifted (the config also excludes `/preview`, the shared helper doesn't).
- The icon-only author link in `Quote.astro` (no name, just the link icon) is missing `data-umami-event`/`data-umami-event-page`, unlike the name+link variant right above it.
- `isExternal` is now inlined in three components (`Quote.astro`, `CtaStrip.astro`, `GrantFaqSection.astro`). Worth pulling into `src/utils/shared/url.ts`.

Everything else looks solid: good CMS test coverage, populate configs cover all four content types, and the author link handling goes through `isSafeMarkdownHref` correctly.

JonathanMatthey:
@Infi-Knight - resolved - please review again / approve

Infi-Knight:
## Re-review

All three points from the last pass are fixed, checked each one:

- **Blockquote double-quoting** (`1bd09dcd`): `formatBlockquote` now only strips marks instead of re-wrapping, and `Blockquote.astro` owns the presentation quotes on both the prop and slot paths. Re-ran my earlier repro (same fixture, same route) and it now renders as a single `"Money should move like data."`, no doubling. Round-trip tests updated to match.
- **`isExternal` duplication + missing analytics** (`9af68e78`): pulled into `isExternalHref` in `src/utils/shared/url.ts` with tests, wired through `Quote`, `CtaStrip`, `GrantFaqSection`. The icon-only author link in `Quote.astro` now also carries `data-umami-event`/`data-umami-event-page`, matching the name+link variant.
- **Demo-path duplication** (`ba3eb2be`): `astro.config.mjs` now imports `isDemoPathname` instead of reimplementing the segment match.

Ran the full suite on the branch: CMS tests (906 passed), root tests (235 passed), lint clean.

Nothing else standing out. Good to merge from my side.

Infi-Knight (approved):

PR #455, opened 2026-07-31
feat(blocks): add agenda component [INTORG-917]

JonathanMatthey:
## Summary

- Add localized Agenda blocks to Foundation, Summit, and Hackathon pages.
- Support Strapi editing, previews, validation, and MDX round trips.
- Add accessible, responsive timeline styling for desktop and mobile.

## Related Issue

Fixes INTORG-917

## Manual Test

1. Add an Agenda block with at least two items in Strapi.
2. Confirm collapsed items display their time.
3. Preview the page at desktop and mobile widths.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] PR title follows Conventional Commits
- [x] Linked issue included
- [x] Scope is focused
- [x] Screenshots for UI changes

PR #463, opened 2026-08-05
fix(strapi): stop split-layout and carousel dropping from foundation, summit, hackathon MDX export [INTORG-1013]

Infi-Knight:
## Summary

`contentPopulate.ts` builds the populate map for Strapi's dynamic-zone queries, but foundation, summit, and hackathon pages were missing `blocks.split-layout` and `blocks.carousel` in that map. A dynamic-zone query only returns components listed there, so those blocks got silently dropped from the MDX export on every Publish, even though they stayed fully visible in the admin form the whole time.

PR #455 (agenda component) added those same entries as a side effect while this branch was in flight, so most of that part of the diff here is now just a doc-comment update and a duplicate-entry cleanup in `GRANT_PAGE_CONTENT_POPULATE`.

The part still needed: once the blocks stop being dropped, LogoCarousel names came back empty, because `updateMediaAlt` (which saves each logo's name as the upload's alt text on import) was only wired up for grant pages. Added `createMediaAltUpdater` in `mdxTransformer.ts` and wired it into the foundation/summit/hackathon import paths in `config.ts`.

Also added `contentPopulate.test.ts`, which walks each content type's schema and checks that every allowed component is in its populate map, so this doesn't get missed a third time.

## Related Issue

Fixes INTORG-1013

## Manual Test

Verified against a local Strapi instance for all three page types:

- Reimported the demo pages with `sync:mdx --force`; partner-logo uploads now carry alt text.
- Edited an unrelated field on each demo page in the admin and clicked Publish.
- Confirmed the exported MDX kept every block and that carousel names round-tripped correctly.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused (6 files)
- [ ] Screenshots for UI changes (N/A, no UI changes)

### DIFF
--- a/cms/scripts/sync-mdx/config.test.ts
+++ b/cms/scripts/sync-mdx/config.test.ts
@@ -1,0 +1,98 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { buildContentTypes, type ContentTypes } from './config'
+import type { StrapiClient } from './strapiClient'
+import { createMdxFile } from './test-utils'
+
+const LOGO_URL = '/img/partner-logos/covenant.avif'
+const LOGO_UPLOAD_ID = 42
+
+/**
+ * LogoCarousel keeps each logo's name on the upload's `alternativeText`, so a
+ * content type that allows `blocks.carousel` has to hand the parser an
+ * `updateMediaAlt` callback. Content types are the only place that wiring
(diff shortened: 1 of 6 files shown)


JonathanMatthey (approved):
Solid, focused fix. 

i like createMediaAltUpdater

PR #465, opened 2026-08-06
fix(cms): parse pasted markdown into real content in the CKEditor

Infi-Knight:
## Summary

Pasting a hyphenated list (`- a`, `- b`, `- c`) into a CKEditor rich-text field and then clicking the bulleted-list toolbar button did not produce a real list. Autoformat only converts `- ` / `1. ` into a list when those characters are *typed*, not when text already containing them is pasted, so the pasted lines stayed as literal paragraph text (`- a`, `- b`, `- c`). Clicking the toolbar's list button afterward just wrapped that literal text in `<li>` tags instead of producing clean list items, leaving stray `-` characters behind (or losing items if the surrounding markup broke, as tracked separately in INTORG-1025's other root cause, already fixed on `playground`).

Adds `PasteFromMarkdownExperimental` (from `@ckeditor/ckeditor5-markdown-gfm`, already an installed peer dependency of the CKEditor plugin, matching how `cms/src/admin/app.tsx` already imports `@ckeditor/ckeditor5-heading` and `ckeditor5` directly) to the shared markdown preset's `extraPlugins`. It intercepts the clipboard pipeline and parses pasted plain-text/markdown content into real editor content, so a pasted list becomes an actual list immediately, no toolbar click needed.

## Related Issue

INTORG-1025

## Manual Test

1. Open any field using the "Basic Markdown" or "Default Markdown" CKEditor preset (e.g. a Title Card description).
2. Paste a plain-text hyphenated list, e.g.:
   ```
   - one
   - two
   - three
   ```
3. Confirm it renders as a real bulleted list immediately after pasting, with no leftover `-` characters and no toolbar click required.
4. Confirm typing `- ` still autoformats into a list as before (unchanged behavior).
5. Confirm pasting rich HTML (e.g. from a Google Doc) is unaffected.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

### DIFF
--- a/cms/src/admin/app.tsx
+++ b/cms/src/admin/app.tsx
@@ -4,6 +4,7 @@ import {
 } from '@_sh/strapi-plugin-ckeditor'
 import type { PluginConfig, Preset } from '@_sh/strapi-plugin-ckeditor'
 import type { HeadingOption } from '@ckeditor/ckeditor5-heading'
+import { PasteFromMarkdownExperimental } from '@ckeditor/ckeditor5-markdown-gfm'
 import type { Editor } from 'ckeditor5'
 import { formatSplitLayoutPanelTitle } from '../plugins/split-layout-type-picker/admin/layoutTypeLabels'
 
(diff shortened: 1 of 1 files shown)


JonathanMatthey (approved):
lovely and nice and short,

nice use of their already built fn

PR #466, opened 2026-08-06
ci: cache key will now be based on content hash instead of run id

bosbaber:
## Summary
In previous iterations the pipeline would store the new set of optimised images into the cache using a key based on the run id of the pipeline. This PR changes this mechanism to rather store the newly generated set of optimised images into the cache against a key based on the hash of the content used during optimisation.

This change should not only improve warmup times but would also allow us to better utilise the github cache.

However, this change should be viewed as an intermediate step and is not going to solve the growing problem of image optimisation taking longer and longer, and having to be repeated each time a single image was added. This will be addressed in an upcoming PR.

- Introduces a new workflow so that we can run the image optimisation manually
- Moved the optimisation calling to a centralised github action for less duplication
- Added cache warming

image

## Manual Test
A reviewer can see the actions runs associated with this PR.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

---

## AI description
This pull request introduces a content-addressed caching system for optimized images in the GitHub Actions workflows, significantly improving CI performance and reliability. It does this by adding a reusable composite action for caching optimized images and updating the workflows to use this action. Additionally, a new job is added to pre-warm the cache on branch builds, ensuring that pull requests can restore optimized images efficiently and avoid long cold-encode times.

**Optimized image caching improvements:**

* Added a new composite action `.github/actions/cache-optimized-images` that restores and saves `public/img/optimized` using a content-based cache key, ensuring exact cache hits when the image sources or encoder change. This avoids unnecessary cache misses and redundant encodes.
* Updated the `pr-checks.yml` workflow to use the new `cache-optimized-images` action, replacing the previous run-id-based cache key with a content-addressed key for more efficient and reliable caching.

**Workflow enhancements:**

* Added a new unconditional `warm-image-cache` job to `merge.yml` that pre-populates the cache for the current branch, so pull requests targeting this branch can restore the cache and avoid expensive cold encodes. This job is deliberately unconditional to keep cache entries fresh and prevent eviction.
* Updated documentation and comments in `merge.yml` to explain the cache warming strategy and its benefits for PR builds.
* Added `warm-image-cache` as a dependency of the `notify-slack` job in `merge.yml` to ensure cache warming failures are surfaced and not silently ignored.

### DIFF
--- a/.github/actions/cache-optimized-images/action.yml
+++ b/.github/actions/cache-optimized-images/action.yml
@@ -1,0 +1,44 @@
+name: Cache optimized images
+description: >
+  Restores and saves public/img/optimized, keyed by the content of the image
+  sources and of the encoder itself.
+
+# Why this is a shared action rather than an inline `actions/cache` step:
+# pr-checks.yml and merge.yml must produce a byte-identical key, otherwise a PR
+# never matches the cache warmed on its base branch and silently pays a full
+# cold encode (~17 min) instead of restoring. Duplicating the hashFiles()
+# expression across two files makes that failure both easy to cause and
+# invisible when it happens.
+#
+# The key covers two kinds of input:
+#
+#   public/img/**, public/uploads/img/original/**
+#     The sources. Deliberately broad rather than a list of raster extensions:
+#     `**` also matches uppercase extensions (the encoder lowercases before
+#     testing, so IMAGE.JPG is processed) and cannot silently miss a new format.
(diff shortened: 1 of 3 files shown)


JonathanMatthey (approved):
looks good, we'll haev to see it in action to see if it's working 

nice work :)

PR #470, opened 2026-08-06
ci(images): serve images via Netlify Image CDN and skip optimisation during pipeline

bosbaber:
We've made the discovery that Netlify CDN has the ability to perform CDN transformations at the edge and on demand. 

This means that we do not need to perform the image optimisations as a separate script anymore.

This PR proposes the following changes
- Removes the optimisations from the general PR check pipeline
- Ensures the control header is set correctly to control caching appropriately
- Modifies Astro so that it will rewrite the image urls so that it will pull from the CDN instead

Images will now be referenced like this when deployed to Netlify:
`/.netlify/images?url=%2Fimg%2Fhomepage%2Fstefan-thomas.webp&fm=avif&w=640&q=85 640w`

image

- Developers wanting to test with optimised images in their local environments can still run the optimiser locally.

---

## Auto generated comments

This pull request introduces a major shift to using the Netlify Image CDN for image optimization, eliminating the need to pre-generate and cache optimized image variants during builds. The changes streamline CI/CD workflows, reduce build times, and ensure that image handling is consistent and efficient across environments. Key updates include removing the image optimization cache from GitHub Actions, conditionally running image optimization scripts, and updating runtime logic to gracefully handle missing or unoptimized images.

**CI/CD Workflow and Caching Updates:**

- Removed the `warm-image-cache` job and all related cache logic from `.github/workflows/merge.yml` and `.github/actions/cache-optimized-images/action.yml` since image variants are no longer pre-generated or cached; images are now optimized on demand via the Netlify Image CDN. [[1]](diffhunk://#diff-fc013cbca44f27bffc398ee0ca938e8d3b035bdd0676b614afed13e01be768b9L1-L48) [[2]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L6) [[3]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L28-L66) [[4]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L383-R343) [[5]](diffhunk://#diff-1da7a27709d11ac16eef3de8e46ed10672b7b1d63a790ffbcbd23a94b19efdcbL60-R67)
- Updated the Netlify build plugin (`netlify/plugins/cache-images/index.js`) to skip restoring or saving the optimized image cache when the CDN is enabled, preventing unnecessary bloat in deploys. [[1]](diffhunk://#diff-9a4e4ae3b7702af33ee0a1e74177a175d2d32caf7b4473b907dd3ec8a0691238R3-R28) [[2]](diffhunk://#diff-9a4e4ae3b7702af33ee0a1e74177a175d2d32caf7b4473b907dd3ec8a0691238R40-R46)

**Image Optimization Pipeline and Runtime Handling:**

- Refactored the image optimization script (`scripts/optimize-images.ts`) to check if the CDN is enabled and, if so, skip encoding images and only generate a catalog of available sources for runtime use. This ensures that the build process is fast and only ships necessary metadata. [[1]](diffhunk://#diff-ca2187bf71bc55d2241962e74d558278dd1ba59d15355a10fd63de612b62a89fR6-R16) [[2]](diffhunk://#diff-ca2187bf71bc55d2241962e74d558278dd1ba59d15355a10fd63de612b62a89fR33-R41) [[3]](diffhunk://#diff-ca2187bf71bc55d2241962e74d558278dd1ba59d15355a10fd63de612b62a89fR146-R175) [[4]](diffhunk://#diff-ca2187bf71bc55d2241962e74d558278dd1ba59d15355a10fd63de612b62a89fL213-L221) [[5]](diffhunk://#diff-ca2187bf71bc55d2241962e74d558278dd1ba59d15355a10fd63de612b62a89fR285-R312)
- Updated the Astro config (`astro.config.mjs`) to pin the CDN decision at build time and expose it to the frontend, ensuring consistent SSR and runtime behavior. Also, added an audit integration for image optimization. [[1]](diffhunk://#diff-e0f0c5adbe0b9ca5d0b57caf5cea33a8d88899fd02a43df1e9862b185f8a1e5fR14-R16) [[2]](diffhunk://#diff-e0f0c5adbe0b9ca5d0b57caf5cea33a8d88899fd02a43df1e9862b185f8a1e5fR51) [[3]](diffhunk://#diff-e0f0c5adbe0b9ca5d0b57caf5cea33a8d88899fd02a43df1e9862b185f8a1e5fR151-R156)

**Frontend and User Experience Improvements:**

- Enhanced the `ImageBlock.astro` component to detect and warn when an image falls back to the unoptimized original (e.g., if a CMS upload is missing from the current deploy), surfacing this via a `data-unoptimized-src` attribute and a console warning. [[1]](diffhunk://#diff-ed153d8f252f5504f1cdb7e42da20cfbfca2d9ad17cc1125079fab911266386cL2-R6) [[2]](diffhunk://#diff-ed153d8f252f5504f1cdb7e42da20cfbfca2d9ad17cc1125079fab911266386cR31-R35) [[3]](diffhunk://#diff-ed153d8f252f5504f1cdb7e42da20cfbfca2d9ad17cc1125079fab911266386cR54) [[4]](diffhunk://#diff-ed153d8f252f5504f1cdb7e42da20cfbfca2d9ad17cc1125079fab911266386cR87-R102)
- Began using a shared `OptimizedImage` component in `BlogPostHeader.astro` for consistency.

**Netlify Configuration:**

- Reordered and clarified Netlify header rules in `netlify.toml` to ensure correct cache-control behavior for all assets, especially for the CDN's transformed images. [[1]](diffhunk://#diff-ab8f79b68b7adff7a07db953bf453f3c5aa6ade98d2b1b67d8432b36392489edR56-R73) [[2]](diffhunk://#diff-ab8f79b68b7adff7a07db953bf453f3c5aa6ade98d2b1b67d8432b36392489edL115-L120)

These changes collectively modernize image handling, reduce CI/CD resource usage, and improve site reliability and observability regarding image delivery.


src/utils/main/imageCdn.ts
```diff
@@ -0,0 +1,102 @@
...
+  sourcePath: string,
+  { format, width }: ImageCdnUrlOptions
+): string {
+  const query = new URLSearchParams({
+    url: sourcePath,
+    fm: format,
+    w: String(width),
+    q: String(QUALITY_BY_FORMAT[format])
+  })
+  return `${NETLIFY_IMAGE_ENDPOINT}?${query}`
```
Infi-Knight, line 83: This comes back with `url` already percent-encoded, which breaks `getHeroSectionStyle` in `src/utils/main/heroSectionStyle.ts`. It does `encodeURI(fullSrc ?? trimmed)`, and `encodeURI` doesn't treat `%` as an existing escape, so we get:

```
url('/.netlify/images?url=%252Fuploads%252Fimg%252Foriginal%252Fgrant_overview_hero_52b51629b7.png&fm=webp&w=3840&q=90')
```

Netlify decodes `url` to the literal string `%2Fuploads%2F...png`, which isn't a path, so the hero background 404s and the section renders with no image. I reproduced it on the branch with the grant overview hero.

It was safe before because `fullSrc` was `/img/optimized/...-full.webp` with nothing to re-encode. Callers are `FoundationContentPage.astro:57` and `SummitContentPage.astro:57`, both reachable through `/page-preview` where `heroImage` comes straight from Strapi.

Simplest fix is to only `encodeURI` in the non-CDN branch, since a CDN URL is already CSS-safe.

src/utils/main/images.ts
```diff
@@ -144,13 +186,19 @@ function listSizedVariants(base: string, ext: 'webp' | 'avif'): ImageVariant[] {
...
  */
 export function getOptimizedImage(src: string): OptimizedImage {
-  const base = getOptimizedBase(src)
-  if (!base) {
+  const source = resolveOptimizableSource(src)
+  if (!source) {
     return { variants: [], fullSrc: null, avifVariants: [], avifFullSrc: null }
   }
 
+  if (imageCdnOverride ?? isImageCdnEnabled()) return buildCdnImage(source)
```
Infi-Knight: Returning early into `buildCdnImage` also drops the only "does this image actually exist" check we had. `listSizedVariants` / `optimizedVariantExists` used to consult the catalog, so a source with no variants returned the all-null `OptimizedImage`, and `OptimizedImage.astro:144` / `ResponsiveSources.astro:46` fell through to a plain ``.

The case that worries me: a Strapi asset at `https://cms.../uploads/img/original/x.png` rendered on `/page-preview`. `resolveOptimizableSource` strips the origin, so the CDN source becomes `/uploads/img/original/x.png` on our own domain. Uploads only reach the repo through the async git auto-commit in `cms/src/index.ts`, so between the upload and the next deploy that path isn't in `dist`. The `<source>` 404s, and browsers do not fall back to `` once they've picked a `<picture>` source, so the image just vanishes. The same input used to produce a working absolute CMS URL.

Worth keeping some existence signal here, or scoping the CDN branch to `/img/**` and leaving `/uploads/**` on the catalog path.

scripts/optimize-images.ts
```diff
@@ -210,6 +208,20 @@ async function processImage(
...
+  // would be ~17 minutes of build time producing files nothing references.
+  // getOptimizedImage() switches to /.netlify/images on the same signal, and
+  // the runtime catalog falls back to its committed stub when absent.
+  if (isImageCdnEnabled()) {
+    console.log(
+      'Netlify Image CDN is enabled — skipping image optimization.\n' +
+        'Images are transformed on demand at /.netlify/images.\n' +
+        'Set IMAGE_CDN=off to force build-time encoding.'
+    )
+    return
```
Infi-Knight: This return also skips the oversized-image guard. The `isImageOverSizeLimit` scan and its `throw` sit further down `main()` (around 240-255), so in CDN mode they never run, and `pr-checks.yml` now sets `IMAGE_CDN=on`. Net effect: someone commits a 15MB PNG under `public/img/`, PR checks stay green, and it ships. The size checks in `cms/src/index.ts` only cover admin uploads, not files committed directly to the repo.

The scan is just `fs.statSync`, no encoding involved, so hoisting it above this early return keeps the guard for free.

src/utils/main/imageCdn.ts
```diff
@@ -0,0 +1,102 @@
...
+ * using pre-generated variants — `/.netlify/images` does not exist off-platform
+ * and would 404 every image.
+ */
+export function isImageCdnEnabled(
+  env: Record<string, string | undefined> = process.env
+): boolean {
+  const override = env.IMAGE_CDN?.trim().toLowerCase()
+  if (override === 'on') return true
+  if (override === 'off') return false
+  return isSet(env.NETLIFY)
```
Infi-Knight, line 50: Reading `process.env` at call time means this is re-evaluated per request on the SSR routes rather than once per build. `page-preview.astro`, `preview/page-preview.astro`, `profile-preview.astro`, `blog/preview.astro` and `developers/roadmap.astro` are all `prerender = false`.

`NETLIFY` is documented as build metadata. If it isn't injected into the functions runtime, those routes take the pre-generated branch and read the committed empty stub (the real manifest is gitignored and never written in CDN mode), so every preview image silently degrades to the unoptimized original.

Did you get a chance to check what `process.env.NETLIFY` looks like inside a deployed function? If it's absent there, pinning the decision at build time via a Vite `define` would be safer than resolving it per request.

Infi-Knight (commented):
Went through this against `staging`. The direction is right and the `resolveOptimizableSource` / `getOptimizedBase` split is a genuine improvement, tests pass on the branch (36/36 across `imageCdn.test.ts` and `images.test.ts`).

One thing will break on deploy: hero backgrounds get double-encoded and 404. Details inline on `imageCdn.ts`. The other blocker-ish one is that `pr-checks.yml` setting `IMAGE_CDN=on` means we lose the oversized-image guard in CI.

Rest of the comments are about the CDN branch no longer consulting the catalog, which quietly removes the "does this file exist" fallback that several components lean on.

Things I checked and think are fine: the `netlify.toml` header reordering, `URLSearchParams` encoding commas so CDN URLs are safe inside `srcset`, no leftover references to `/img/optimized/*` in content or workflows, and the `cache-images` plugin no-op guard mirroring `isImageCdnEnabled` correctly.

src/utils/main/imageCdn.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/** One transformation URL per target width, ascending. */
+export function buildImageCdnVariants(
+  sourcePath: string,
+  format: ImageCdnFormat,
+  widths: readonly number[] = TARGET_WIDTHS
+): ImageCdnVariant[] {
+  return widths.map((width) => ({
+    src: buildImageCdnUrl(sourcePath, { format, width }),
```
Infi-Knight, line 98: Every image now advertises 640w through 3840w regardless of its intrinsic width. Netlify clamps so nothing renders wrong, but take a 300px author avatar: all five candidates return the same 300px bytes. Across a real audience with varying DPR and `sizes`, that's up to five billed transforms and five edge-cache entries where the encoder used to emit one exact-width file.

It's the same "second billed transform for the same image" concern the `buildImageCdnUrl` doc comment raises for unsized URLs, just five times over. Might be worth capping the default list and letting the hero opt into 2560/3840, since it's about the only thing that needs them.

src/utils/main/images.test.ts
```diff
@@ -138,3 +150,110 @@ describe('getOptimizedImage', () => {
...
+    }
+  })
+
+  it('does not re-transform an already-optimized path', () => {
+    setImageCdnEnabledForTests(true)
+
+    expect(getOptimizedImage('/img/optimized/hero-640.webp')).toEqual(EMPTY)
+  })
+
+  it('reports variants as available so components render a <picture>', () => {
```
Infi-Knight, line 344: This is correct for the components as they stand, but it does turn a real check into a constant. `HomepageHero.astro:32` uses `hasOptimizedVariants(getOptimizedImage(HERO_HIGHRES_SRC))` as a "did the encoder actually produce this file?" probe, and in CDN mode it's now always true.

So if `stefan-thomas-highres.avif` ever gets renamed or removed, we still emit the >=2560px `<source>` and 4K displays get a broken hero instead of quietly falling back to the standard image.

src/components/pages/ProfileDetailPage.astro
```diff
@@ -51,15 +51,18 @@ const colorIndex = await getProfileColorIndex(
...
 
-const avatarSrc = photo
+// `isAvif` is tracked rather than sniffed from the URL: a Netlify Image CDN URL
+// carries its format in a query parameter, not a file extension, so an
+// endsWith('.avif') check would silently drop the preload's type hint.
+const avatar = photo
   ? (() => {
       const { avifFullSrc, fullSrc } = getOptimizedImage(photo)
-      return avifFullSrc ?? fullSrc ?? photo
+      if (avifFullSrc) return { src: avifFullSrc, isAvif: true }
```
Infi-Knight: In CDN mode `avifFullSrc` is never null, so this branch always wins and every avatar renders as a bare `` with an `image/avif` preload and nothing behind it. Previously `avifFullSrc` was null whenever no AVIF variant existed (a GIF profile photo, or anything the encoder skipped) and we fell back to `fullSrc ?? photo`.

The `isAvif` tracking is the right fix for the old `endsWith('.avif')` sniff. The gap is that there's no `<picture>` with a webp fallback wrapping it, so a client without AVIF decode support now gets no avatar at all.

src/utils/main/imagePaths.ts
```diff
@@ -1,5 +1,23 @@
 export const TARGET_WIDTHS = [640, 1280, 1920, 2560, 3840] as const
 
+/**
+ * Encoding quality, shared by the two things that can produce a variant: the
+ * build-time encoder (`scripts/optimize-images.ts`) and the Netlify Image CDN
+ * URL builder (`imageCdn.ts`). Defined here so the two cannot drift, and so a
+ * change to either lands in the CI cache key (this file is hashed by
+ * `.github/actions/cache-optimized-images`).
```
Infi-Knight, line 17: This points at `.github/actions/cache-optimized-images`, which the PR deletes. `git grep` says it's the last reference left. The constants still belong here for the drift reason above, just needs the cache-key half dropped.

src/utils/main/imageCdn.ts
bosbaber: Thank you, I applied the suggested fix. Clearly there are a lot of details about this codebase I need to still learn. I also added a test to ensure regression.

src/utils/main/images.ts
bosbaber: - Just to make sure it is not obvious, we should not be having references back to the strapi site since it will be firewalled in the future
- The bulk of images that will be served will be from the uploads folder so we definitely need to have them on the cdn
- I made the code more sophisticated so that the build tries to track what should be available. If it cannot be found then we will serve full size src instead

src/utils/main/imageCdn.ts
bosbaber: Reworked to use a vite define as suggested.

src/utils/main/imageCdn.ts
bosbaber: Capped the default width with a ladder of available sizes. See the changes in the buildCdnImage function that tries to address this issue.

scripts/optimize-images.ts
bosbaber: Fixed this issue by making these checks happen first.

src/components/pages/ProfileDetailPage.astro
bosbaber: I believe this should now be resolved as well. Changed this component to use the standard machinary which is supposed to deal with this automatically.

bosbaber: `ProfileDetailPage.astro` will now delegate the heavy lifting to `OptimizedImage`, which will the appropriate URLs pointing to the optimised images.

src/components/shared/OptimizedImage.astro
```diff
@@ -152,6 +175,23 @@ const imgSrc = media ? undefined : src
       decoding={decoding}
       fetchpriority={fetchpriority}
       draggable={draggable}
+      data-unoptimized-src={src && isOptimizableSource(src) ? src : undefined}
```
Infi-Knight: The fallback `` branch never applies `pictureClass`, only `class`.

This PR moves `BlogPostHeader`'s breakpoint toggles into that prop (`tablet:hidden`, `hidden tablet:block`). I rendered a post whose feature images are not in the catalog, and both come out as:

```html
<img ... class="w-full h-auto rounded-3xl">
```

Both visible at every breakpoint, stacked. Fix is to merge `pictureClassName` into this class list.

Not urgent: no post sets `featureImageMobile` yet. It is a live Strapi field though, so one content edit reaches it.

src/integrations/audit-image-optimization.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (src.startsWith('/img/optimized/')) return false
+  return /\.(png|jpe?g|webp|avif)$/i.test(stripQuery(src))
+}
+
+/** Classifies a single `<img>` tag, or `null` when it is fine. */
+function auditImgTag(tag: string): ImageAuditViolation | null {
+  if (hasAttr(tag, 'data-allow-unoptimized')) return null
+
+  const degraded = getAttr(tag, 'data-unoptimized-src')
+  if (degraded !== null) {
```
Infi-Knight, line 155: This makes the graceful degrade a failed deploy.

`data-unoptimized-src` is emitted when a source is optimizable but missing from the catalog. `imagePaths.ts:48` calls that the intended outcome for an upload not yet git-synced from the CMS. Throwing on it means one un-synced image aborts the whole build.

Suggest logging `degraded-marker` and failing only on `standalone-raw` and `picture-without-cdn`. Those two mean a component bypassed `OptimizedImage`, which is a code defect and worth blocking. A missing source is content timing, and already has a designed fallback.

For context: `IMAGE_CDN=on pnpm run build` passes today, so this is a latent path, not a current break.

src/utils/main/images.ts
```diff
@@ -69,13 +158,22 @@ function replaceUrlPathPrefix(
...
+ * Absolute Strapi URLs are reduced to their site-relative pathname: the CMS is
+ * firewalled and the site must stay self-contained, so every source has to
+ * resolve against our own deploy, never the CMS origin.
  */
-function getOptimizedBase(src: string): string | null {
-  if (!src || src.endsWith('.svg')) return null
+function resolveOptimizableSource(src: string): ResolvedImageSource | null {
+  // GIFs are excluded like SVGs: encoding to WebP/AVIF (or a CDN fm= transform)
+  // would drop animation, so they ship as-is (mirrors scripts/optimize-images.ts).
+  if (!src || /\.(svg|gif)$/i.test(src)) return null
```
Infi-Knight: Denylist here (`svg|gif`), allowlist in the catalog builder (`RASTER_EXTENSIONS` in `optimize-images.ts:47`, jpg/jpeg/png/webp/avif).

So `/img/photo.jfif` is optimizable here but can never be in the catalog. I checked, it returns the empty result and degrades permanently. With the audit above that is a build failure fixable only by renaming the file.

One shared extension list across the three places would remove the drift.

src/integrations/audit-image-optimization.ts
```diff
@@ -0,0 +1,193 @@
...
+        // Only meaningful for what Netlify ships: CDN mode delivers via
+        // /.netlify/images. A non-CDN local build uses /img/optimized/* paths
+        // and would misfire here.
+        if (!isImageCdnEnabled()) {
+          logger.info('CDN off — skipping image optimization audit.')
+          return
+        }
+
+        const distDir = fileURLToPath(dir)
+        const files = await collectHtmlFiles(distDir)
```
Infi-Knight, line 270: Nothing checks that this scanned anything. If `collectHtmlFiles` returns nothing, the hook logs success and passes.

Cheap insurance: log the file count, throw when it is 0.

Also worth noting `audit-image-optimization.test.ts` covers the two pure helpers but never calls `auditImageOptimization()` itself.

src/utils/main/images.ts
```diff
@@ -164,3 +307,68 @@ export function getOptimizedImage(src: string): OptimizedImage {
...
+  const source = resolveOptimizableSource(src)
+  if (!source) return image
+
+  const rawFormat =
+    DELIVERABLE_SOURCE_FORMATS[path.extname(source.pathname).toLowerCase()]
+
+  const rungFor = (format: ImageCdnFormat): ImageVariant => ({
+    src:
+      rawFormat === format
+        ? source.pathname
```
Infi-Knight: The raw-source rung goes into a srcset unencoded, so a space in the filename breaks it:

```
... &w=1920&q=85 1920w, /img/hero image.avif 2000w
```

The browser reads `/img/hero` as the URL and `image.avif` as the descriptor. The CDN rungs are safe because `URLSearchParams` encodes them; this one is not. `heroSectionStyle.test.ts` already uses a spaced filename as a fixture.

Infi-Knight (commented):
Second pass. All eight points from the first round are addressed: the oversize guard runs before the CDN early return, the deployed-sources catalog restores the existence check, the Vite define pins the CDN decision at build time, `heroSectionStyle` no longer double-encodes, `ProfileDetailPage` delegates to `OptimizedImage`, and the width ladder plus `withIntrinsicWidthRung` fix the clamped-rung waste. Only the stale doc comment in `imagePaths.ts` is left.

I verified this against a real `IMAGE_CDN=on pnpm run build` and the deploy preview, not just by reading. The build passes and the audit reports clean. For contrast, the same check over a pre-PR `dist` flagged 64 blog feature images as raw, so the `BlogPostHeader` conversion is doing real work.

The hero path checks out too. Above 2560px the AVIF source is served as the file itself, no transform and no re-encode, which is exactly what `withIntrinsicWidthRung` promises. I confirmed the 3463 constant against the asset.

Comments below. The `pictureClass` one is the only real bug; the rest are small.

src/components/blocks/ImageBlock.astro
```diff
@@ -74,3 +84,19 @@ const TABLET_MQ = '(max-width: 1199px)'
     )
   }
 </figure>
+
+<script>
```
Infi-Knight, line 88: Duplicated verbatim in `OptimizedImage.astro`. It ships client JS on nearly every page to `console.warn` about something the build-time audit already catches, and puts internal source paths into production HTML. Suggest gating on `import.meta.env.DEV` or dropping it.

src/integrations/audit-image-optimization.ts
```diff
@@ -0,0 +1,193 @@
...
+ * is a violation. Escape hatch: `data-allow-unoptimized` on the `<img>`.
+ */
+import type { AstroIntegration } from 'astro'
+import { readdir, readFile } from 'node:fs/promises'
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { isImageCdnEnabled } from '../utils/main/imageCdn'
+
+const CDN_MARKER = '/.netlify/images'
+const IMG_TAG_RE = /<img\b[^>]*>/gi
```
Infi-Knight, line 53: The audit only looks at `` and `<picture>`. CSS backgrounds and `<video poster>` pass silently.

`AnimatedNetworkProblem.astro:51` serves `/img/home/video-problem-poster.jpg` (60KB) both ways. It is in the catalog, so the CDN could transform it, but nothing routes it there and the build still reports clean.

Not a blocker. Either widen the audit to `background-image:url(...)` and `poster=`, or soften the success message so it does not read as a whole-site guarantee.

Infi-Knight (commented):
One more, on a surface the audit does not cover.

Cleaned up my earlier comments: I withdrew two after checking them against the deploy preview. The netlify.toml caching one was wrong, your comment is correct and transformed images do inherit `max-age=86400` from the source. The `heroSectionStyle` one does not apply to any page that exists today.

Infi-Knight:
To save you weighing all seven equally:

**Blocking: one.** The `degraded-marker` throw in `audit-image-optimization.ts:65`. `CLAUDE.md:129` in this same PR calls the missing-upload window "the expected degrade-to-`` window, not a bug", and the audit aborts the build on exactly that. Worth settling either way before merge. If strict is what you want, the CLAUDE.md line should change instead.

**Worth doing here: one.** The `pictureClass` drop in `OptimizedImage.astro:178`. One line, and this PR is what created the dependency on it.

**Fine as follow-ups: the rest.** The `.jfif` mismatch, the spaced srcset, the zero-file guard, the CSS background gap, and the duplicated warn script. None of them can fire on current content.

For the record, the PR works. `IMAGE_CDN=on pnpm run build` passes, the preview is green, the hero serves the AVIF source directly above 2560px with no re-encode, and the `BlogPostHeader` conversion cleaned up 64 blog feature images that were being served raw before.

src/components/shared/OptimizedImage.astro
bosbaber: Hardened the behaviour of OptimizeImage as suggested.

src/integrations/audit-image-optimization.ts
bosbaber: - I take the point and therefore I made the `degraded-marker` scenario a warning instead.

src/components/blocks/ImageBlock.astro
bosbaber: added proposed gates

src/integrations/audit-image-optimization.ts
bosbaber: Sorted

src/utils/main/images.ts
bosbaber: Fixed.

Infi-Knight:
Two follow-ups, neither blocking.

`auditTagAttributes` has no `degraded-marker` equivalent, so `raw-css-background` and `raw-poster` fail the build on exactly the degrade path `` is now exempt from. `VideoEmbed.astro:27` and `heroSectionStyle.ts:10` both emit `fullSrc ?? rawPath`, so a missing upload behind either one aborts the deploy. Nothing reaches it today: the only poster is a committed `/img/` file, and `getHeroSectionStyle` has no rendering callers. Setting `data-allow-unoptimized` in those two fallback branches would be enough.

`resolveOptimizableSource` decodes only inside the `http` branch, so `/uploads/img/original/hero%20image.jpg` misses the catalog while the absolute CMS URL for the same file matches.

Infi-Knight (approved):

PR #422, opened 2026-07-23
refactor(strapi,astro): add image alternativeText localized support consistent across components

adrianboros:
### Problem
Strapi's Media Library stores alternativeText/caption on the upload file itself — a single global value shared across every locale and every place that image is used. There was no way to give the same image different alt text per locale, or to swap in a translated image (e.g. a graphic with baked-in English text) for a specific locale.
### Solution
Added a new reusable Strapi component, shared.localized-media (image + alternativeText, both pluginOptions.i18n.localized: true), and migrated the components/content-types that carry meaningful content images to use it instead of a bare media field:

- `blocks.image-block` - `image`/`altText` → media component (tablet/mobile variants stay plain media fields, sharing the primary's alt text)
- `blocks.split-layout` - `image` → `media` component (previously had no alt field at all; relied solely on the native upload's global alt)
- `blocks.image-row` - same pattern (currently not wired into any content-type's dynamiczone, so no live data affected)
shared.article-bio - `profileImage` → `media` component
- `foundation-blog-post` - `featureImage`/`thumbnailImage` → `featureMedia`/`thumbnailMedia` components (`featureImageMobile` stays a plain field)
- `profile-page` - `photo` → `media` component
- `shared.hero` - `backgroundImage` → `media` component; `backgroundImageMobile` stays plain and **shares the same alt text as the desktop image** (single `heroImageAlt` field, not a separate mobile one)

**Out of scope**: `shared.hero`'s image stays decorative (`alt=""`) for `foundation-page`/`summit-page`, which render it as a CSS background-image, not an  — no alt attribute applies there. Only `grant-page`/`grant-overview-page` (which render an actual `` via `PageHero.astro`) get `heroImageAlt` in their frontmatter schema.

### Sync pipeline & Astro

- `mdxTransformer.ts` (MDX→Strapi) and `mdx.ts/blogLifecycle.ts/profileMdx.ts` (Strapi→MDX) updated on both directions so `alt`/`imageAlt`/`photoAlt`/`featureImageAlt`/`heroImageAlt` frontmatter fields round-trip correctly through the new nested shape.
- All affected `populate` configs updated (dynamiczone blocks, hero, article bio, profile relations) — component fields aren't populated by Strapi by default, so several were previously under-populated.

- Astro-side live-preview pages (`blog/preview.astro`, `es/blog/preview.astro`, `profile-preview.astro`) and `ProfileBlock.astro/profiles.ts` updated to match the new Strapi response shape.
[...]

PR #473, opened 2026-08-06
fix(cms): restore Media Library alt text for carousel logos [INTORG-1021] INTORG-1054

JonathanMatthey:
## Summary

#422 hid Media Library “Alternative text” globally under the assumption that images use `shared.localized-media`. Logo Carousel still used a plain multi-media `logos` field, so editors had nowhere to set per-logo alt text.

This PR moves carousel logos to a **repeatable `blocks.carousel-logo` component** (`image` + per-entry `alternativeText`), wires MDX ↔ Strapi (handler + serializer), maps `blocks.carousel` → `LogoCarousel` in DynamicZone preview, and keeps Media Library file-alt hide scoped so content-form alt fields stay usable.

Also: remove the unused testimonial `Carousel.astro`, skip incomplete logos in preview instead of 500ing the page, and fix `CarouselLogoEntry` typing so the serializer typechecks.

## Related Issue

Fixes INTORG-1021

## Breaking deploy note (ordering matters)

**Changing `logos` from multi-media to a repeatable component does not migrate existing rows.**

On rebuild, Strapi creates the new component tables empty and leaves the old `files_related_mph` rows (`related_type='blocks.carousel'`, `field='logos'`) pointing at nothing. Content Manager then returns `logos: []` for those pages. Re-saving without re-adding logos fails with `400 Carousel block is missing logos` from `validateContentBlocks`.

### Why a normal merge is not enough

`merge.yml`:

1. Sets `cms_changed` when `cms/` changes → **rebuild-strapi** runs (schema applied, logos stranded).
2. Sets `content_changed` only for `^src/content/.+\.(md|mdx)$` (and some public media paths).
3. This PR does **not** touch MDX, so **`sync-to-strapi` is skipped** after a normal push merge.
4. `pr-checks` dry-runs against an empty `ci-dry-run.db` (creates only), so CI does not catch this.

### Required post-merge steps (staging / playground)

1. **Merge the PR** into `staging` (or `playground`) so Strapi rebuilds with the new component schema.
2. **Immediately re-run “Build and Sync” via `workflow_dispatch`** on that same branch.
   - Dispatch forces `cms_changed`, `content_changed`, and `navigation_changed` to `true`.
   - That runs `sync:mdx`, which rewrites every Logo Carousel from MDX into the new `carousel-logo` shape.
3. Confirm pages that used carousels (e.g. demo foundation / summit / hackathon) show logos again and can save/publish without a 400.
4. **Do not** rely on editors manually re-adding logos; use the forced sync.

Without step 2, editors who open those pages cannot save until they re-add every logo by hand.

### Locales

`logos` is now `localized: true` (the old multi-media field was not). Confirm the forced `sync:mdx` covers non-default locale MDX under `src/content/...` (and any locale-specific carousel content). After sync, spot-check at least one non-`en` locale page that has a carousel if one exists; if a locale only lived in the DB and not in MDX, that locale’s logos will still need a manual fix or an MDX source of truth.

## Manual Test

1. After deploy + forced sync (above): open a page with Logo Carousel in Strapi.
2. Confirm logos are present as `Carousel Logo` entries with image + alternative text.
3. Edit a logo’s alt text; save; publish / export; confirm site `alt` matches.
4. Preview a draft with a missing logo image: page should still render (incomplete logos skipped), not 500.
5. Strapi preview / DynamicZone: Logo Carousel block renders logos, not the old testimonial carousel.

## Reviewer notes

- `Carousel.astro` (testimonial quotes) deleted — no importers after DynamicZone maps `blocks.carousel` → `LogoCarousel`.
- Serializer kept: Strapi → MDX on save + `validateContentBlocks`.


cms/src/components/blocks/carousel.json
```diff
@@ -25,10 +25,16 @@
       }
     },
     "logos": {
```
Infi-Knight, line 27: Changing `logos` from a media field to a repeatable component strands the existing logo rows, and I don't think anything puts them back.

I booted this branch against a copy of the dev DB (4 carousel rows holding 7/30/30/30 logos). Strapi created `components_blocks_carousel_logos` and `components_blocks_carousels_cmps` empty, and left all 97 rows in `files_related_mph` (`related_type='blocks.carousel'`, `field='logos'`) pointing at nothing. The content-manager API returns `logos: []` for those pages afterwards.

That much is recoverable. What isn't: re-saving one of those pages *unchanged* fails with `400 Carousel block is missing logos` from validateContentBlocks. So an editor who opens the demo hackathon page can't save or publish it until they manually re-add 30 logos.

The deploy doesn't heal it either. `merge.yml` runs `rebuild-strapi` on `cms_changed`, but gates `sync-to-strapi` on `content_changed`, which `detect-changes` only sets for `^src/content/.+\.(md|mdx)$`. This branch touches no MDX, so the sync that would rewrite logos in the new shape gets skipped. CI misses it too, since pr-checks dry-runs against an empty `ci-dry-run.db` where every entry is a create.

Merging via `workflow_dispatch` would force `content_changed` true and let `sync:mdx` re-push everything, which is probably the cheapest route. Worth confirming that covers non-default locales, though, because `logos` is now `localized: true` where the old media field wasn't. Whichever way you go, can it go in the PR description? Whoever deploys needs to know the ordering matters.

Infi-Knight (commented):
Ran this against a copy of the dev DB. The feature works: per-logo alt saves, the fallback to Media Library alt works, and the exported MDX comes out right. Two things I'd want resolved before this merges (existing carousel data, and a type error in the serializer), plus some smaller notes inline.

cms/src/serializers/blocks/carousel.serializer.ts
```diff
@@ -1,13 +1,59 @@
...
+function logoToMdxItem(logo: CarouselLogoEntry): { name: string; src: string } {
+  // Legacy multi-media entry
+  if ('url' in logo && typeof logo.url === 'string') {
+    return {
+      name: logo.alternativeText ?? '',
+      src: logo.url
+    }
+  }
+
+  const image = logo.image
```
Infi-Knight, line 36: This doesn't compile:

```
src/serializers/blocks/carousel.serializer.ts(38,22): error TS2339: Property 'image' does not exist on type 'CarouselLogoEntry'.
  Property 'image' does not exist on type '{ id: number; url: string; alternativeText: string; }'.
```

The guard on line 31 is a conjunction, so its negation is a disjunction and TS can't narrow the else branch. The legacy member stays in the union, and `logo.image` is an error on it.

Nothing flags it. eslint and vitest don't typecheck, and Strapi's compile step prints `Found 1 error(s)` then carries on, so `cd cms && pnpm run build` (what pr-checks runs) still exits 0. tsc emits valid JS so runtime is unaffected, but the union isn't discriminating anything.

Collapsing it to one interface fixes it:

```ts
interface CarouselLogoEntry {
  image?: MediaField
  alternativeText?: string | null
  /** Legacy shape: plain multi-media entries (pre carousel-logo component) */
  id?: number
  url?: string
}
```

tsc comes back clean with that, and the 23 carousel tests still pass.

src/components/blocks/LogoCarousel.astro
```diff
@@ -1,13 +1,39 @@
...
-const { heading, accessibilityLabel, logos } = Astro.props
+const { heading, accessibilityLabel, logos: rawLogos } = Astro.props
+
+const logos: LogoMdx[] = rawLogos.map((logo) => {
+  if ('src' in logo && typeof logo.src === 'string') {
+    return { name: logo.name ?? '', src: logo.src }
+  }
+  const src = logo.image?.url
+  if (!src) {
+    throw new Error('LogoCarousel logo is missing image url')
```
Infi-Knight: Throwing here takes out the entire preview page instead of one logo.

Only DynamicZone reaches this branch (real pages get the MDX `{name, src}` shape), and validateContentBlocks currently blocks saving a logo row with no image, so I couldn't actually trigger it. It's defensive. But preview is the one surface whose job is rendering half-finished editor state, and having a single bad entry 500 the page with no hint which block did it feels like the wrong trade. Skipping the entry and leaving the CMS-side validation as the gate would be safer, and it lines up with the errors-as-values note in CLAUDE.md.

cms/src/admin/app.tsx
```diff
@@ -204,8 +204,8 @@ export default {
         margin-top: 1.5rem;
         padding-top: 1.5rem;
       }
-      /* TEMP UI Fix: hide Media Library Alternative text — set alt on localized-media per locale instead */
-      div:has(> div > input[name="alternativeText"]) {
+      /* TEMP UI Fix: hide Media Library dialog Alternative text only (not content form fields e.g. carousel-logo) */
```
Infi-Knight: The reasoning in this comment isn't right, although the change itself is harmless.

Strapi names nested component inputs with their full path. The carousel-logo alt field on the content form is `name="content.5.logos.0.alternativeText"` (I checked in the DOM), so the old `div:has(> div > input[name="alternativeText"])` never matched it, being an exact attribute match. Scoping to `[role="dialog"]` doesn't un-hide anything that was hidden before.

I did check it doesn't regress what #422 wanted: opening an asset in the Media Library still gives `visible: false`, `inDialog: true`. So keep the tighter selector if you like it. But as written, the comment and the PR description will both leave the next reader thinking content-form inputs were at risk from unscoped `name=` selectors.

cms/src/serializers/blocks/carousel.serializer.ts
```diff
@@ -1,13 +1,59 @@
...
+  // runs this serializer on the raw write body where `image` is a bare upload
+  // id — empty src is fine there because the output is discarded.
+  const src =
+    typeof image === 'object' && image != null ? (getImageUrl(image) ?? '') : ''
+  const fileAlt =
+    typeof image === 'object' && image != null
+      ? (image.alternativeText ?? '')
+      : ''
+  // Prefer field-level alt on carousel-logo; fall back to file alt for old data
+  const name = logo.alternativeText ?? fileAlt ?? ''
```
Infi-Knight: `fileAlt` is always a string, since both branches of the ternary above return one, so the trailing `?? ''` can never fire. `logo.alternativeText ?? fileAlt` is enough.

src/components/blocks/DynamicZone.astro
```diff
@@ -44,7 +44,7 @@ const componentMap: Record<string, any> = {
   'blocks.paragraph': Paragraph,
   'blocks.cards-grid': CardsGrid,
   'blocks.card-links-grid': CardLinksGrid,
-  'blocks.carousel': Carousel,
+  'blocks.carousel': LogoCarousel,
```
Infi-Knight, line 48: This is a real fix that the PR description doesn't mention. `blocks.carousel` was pointing at `Carousel.astro`, which is the testimonial quote carousel and expects an `items` prop, so previewing any page with a Logo Carousel rendered the wrong component entirely.

One side effect: `Carousel.astro` now has no importers left, just a stale mention in `FaqSectionsNav.astro:60`. Probably fine to delete.

JonathanMatthey:
### alt text in spanish on carousel

image


image

JonathanMatthey:
not on image -> ( which cant be ES )
image

cms/scripts/sync-mdx/profileHandler.ts
```diff
@@ -58,10 +58,17 @@ export function createRelationResolver(
     if (entry instanceof Error) throw entry
     if (entry) return { documentId: entry.documentId }
 
-    if (locale !== 'en') {
+    // Live sync: do not connect EN-only relations into a non-en document —
+    // Strapi relation transforms require the target locale to exist.
+    if (dryRun && locale !== 'en') {
```
Infi-Knight, line 64: Both dry-run escape hatches here are locale-blind, so a dry-run passes on exactly the relation the live sync rejects.

Path 2 fires only when the relation exists in `en` and not in `locale`. Per the comment you added above, that is precisely what Strapi refuses at write time, so there is no input where returning the `en` documentId predicts a successful write.

Path 3 has the same hole, and it is the one that reaches CI. `config.ts:352` builds the set as:

```ts
const profilePathSlugs = new Set(
  scanMDXFiles('profiles', contentTypes).map((f) => f.pathSlug)
)
```

`scanMDXFiles` walks `profiles/` with `readdirSync(recursive: true)`, so `es/` files are included, but `.map` drops `f.locale`. The set holds a pathSlug when any locale has it. `grant/fellowship/caroline-sinders` has an EN file and no ES file, so it sits in the set. Point a new ES page at it and:

- dry-run: path 3 hits, logs "would be created by this same sync run", exits 0
- live sync: throws `UNRESOLVED_RELATION`, because no ES caroline-sinders is ever created

That is the class of failure you just spent the evening on, and the dry-run cannot catch it today. `pr-checks` is the worst case: `ci-dry-run.db` is empty, so path 1 always misses and path 3 is the only branch that can fire.

Keying the set by locale closes both. `syncOperations.ts` already uses this shape for `matchedPathSlugs` (`${localeCode}:${localeMdx.pathSlug}`):

```ts
// config.ts
const profilePathSlugs = new Set(
  scanMDXFiles('profiles', contentTypes).map(
    (f) => `${f.locale || 'en'}:${f.pathSlug}`
  )
)

// profileHandler.ts
if (dryRun && dryRunPathSlugs?.has(`${locale}:${pathSlug}`)) { ... }
```

With that in place path 2 has no remaining purpose and can go. If you want to keep the diagnostic, log the warning and then fall through to the throw instead of returning the EN documentId.

cms/scripts/sync-mdx/carouselHandler.ts
```diff
@@ -65,9 +65,12 @@ async function handleCarousel(
...
       rawLogos.map(async (logo) => {
         const id = await ctx.resolveMediaUpload!(logo.src)
-        // '' means "explicitly no alt text" in Strapi (see carousel.serializer.ts)
-        await ctx.updateMediaAlt?.(id, logo.name || null)
-        return id
+        // name is logo alt in MDX; store on the carousel-logo component field
+        // (not Media Library) so editors set it next to the image in Strapi.
+        return {
+          image: id,
+          alternativeText: logo.name?.trim() ? logo.name : null
```
Infi-Knight: Minor: this tests the trimmed value but stores the raw one, so `name: '  Acme  '` lands in Strapi as `'  Acme  '` and round-trips back into the MDX with the padding still on it. `agendaHandler.ts:84` stores `heading.trim()` for the same pattern.

```ts
alternativeText: logo.name?.trim() ? logo.name.trim() : null
```

Infi-Knight (commented):
Re-reviewed at a992181. All six findings from the previous pass are resolved, and I checked them rather than taking the commits at face value:

- `CarouselLogoEntry` collapsed to one interface: `npx tsc --noEmit` in `cms/` is clean now (was `TS2339` on `logo.image`)
- `LogoCarousel` skips incomplete logos instead of throwing
- the dead `?? ''` after `fileAlt` is gone
- `Carousel.astro` deleted, and the stale `FaqSectionsNav` mention cleaned up
- the `admin/app.tsx` comment no longer claims content-form inputs were at risk
- the migration and forced-`sync:mdx` ordering is documented in the PR description, including the locale caveat

Checks on this branch: 603/603 sync-mdx tests pass, `tsc --noEmit` clean in `cms/`, `pnpm build` exits 0. I also confirmed the feature in the build output: `dist/es/demo-foundation-page` renders 30 visible logos carrying the Spanish per-logo alt text, plus 30 `aria-hidden`/`inert` duplicates with empty alt. All three populate configs and both `page-preview` queries fetch `logos.image`.

The new ES page relations all resolve in ES. Every profile pathSlug referenced by the three ES pages has a file under `src/content/profiles/es/`. `caroline-sinders` is the only EN-only profile, and no ES page references it.

## On `createLocalization`

Separately: `strapiClient.createLocalization` does not need fixing. `PUT /api/:pluralApiId/:documentId?locale=es` is the supported way to create a new locale on Strapi 5.41. The collection-type controller spreads `sanitizedQuery` (which carries `locale`) into the service call, and `document-service/repository.js` `update()` handles the missing-locale case explicitly: it looks up the entry in the requested locale, and when that misses but the document exists in another locale, it runs `copyNonLocalizedFields` and calls `entries.create()`. So a first-time locale create works through that path.

The error was the relation transform, not the locale create. `document-service/transform/relations/transform/data-ids.js:39` throws `Document with id "...", locale "es" not found` when a `connect` target has no row in the source locale. Your 140cf398 fix addresses the real cause, so no follow-up ticket is needed for `createLocalization`.

Three things below. The first is worth fixing before merge, because it is the guard that should have caught this ES bug in the first place.

src/components/blocks/LogoCarousel.astro
Infi-Knight, line 1: Non-blocking, and the markup predates this PR, but mapping `blocks.carousel` here makes it reachable from CMS content, where `heading` is optional (`carousel.json` sets no `required` on it).

With no heading, line 107 renders an empty `<h2 id="logo-carousel-heading">`, and the `aria-labelledby` on the `<section>` then points at an element with no text, so the region ends up with no accessible name. An empty heading also lands in the document outline. The inner `role="group"` still carries `aria-label={accessibilityLabel}`, so the logos themselves stay labelled.

The static `id` is a second-order issue: two carousels on one page give duplicate ids. No MDX file has two today, but a dynamic zone allows it.

Rendering the heading conditionally covers both:

```astro
<section class={sectionClasses} aria-labelledby={heading ? headingId : undefined}>
  {heading && <h2 id={headingId} class={headerClasses}>{heading}</h2>}
```

with `headingId` derived per instance.

Infi-Knight:
Trim fix is right. 603/603 sync-mdx tests still pass.

One thing still open: `profilePathSlugs` (config.ts:352) drops `f.locale`, so a dry-run passes on an ES page pointing at an EN-only profile and the real sync throws. Two-line fix, described in my comment on profileHandler.ts:63.

Infi-Knight (approved):

PR #475, opened 2026-08-06
fix(cms): stop CKEditor duplicated-modules crash in admin

JonathanMatthey:
## Summary

Strapi admin was crashing with `ckeditor-duplicated-modules` because two CKEditor integrations were loading at once, and admin config mixed import paths.

## Changes

- Remove unused official `@ckeditor/strapi-plugin-ckeditor` package
- Import GFM plugin and types only from the unified `ckeditor5` package (with a comment explaining why)
- Pin `ckeditor5` `47.4.0` as a direct dependency so the admin runtime import cannot silently follow a hoisted peer bump
- Enable only the community `ckeditor5` Strapi plugin in `config/plugins.ts`

## Related Issue

CKEditor admin crash (`ckeditor-duplicated-modules`) — please attach Linear ID if not auto-linked.

## Manual Test

1. `cd cms && pnpm install && pnpm develop`
2. Open admin, edit any entry with a CKEditor field (e.g. paragraph / FAQ answer)
3. Confirm admin loads without `ckeditor-duplicated-modules`
4. Confirm paste/markdown still works as before

## Checks

- [x] `pnpm run format` (passed in CI)
- [x] `pnpm run lint` (passed in CI)

## PR Checklist

- [ ] Linked issue included
- [x] Scope is focused

### DIFF
--- a/cms/config/plugins.ts
+++ b/cms/config/plugins.ts
@@ -1,8 +1,15 @@
 import { MAX_MEDIA_BYTES } from '../src/utils/uploadLimits'
 
 export default () => ({
-  ckeditor: {
+  // Community plugin used by all rich-text fields (`plugin::ckeditor5.CKEditor`).
+  ckeditor5: {
     enabled: true
+  },
+  // Official package ships a second CKEditor build under plugin name `ckeditor`.
+  // Leaving it enabled double-loads modules and crashes the admin with
+  // `ckeditor-duplicated-modules`. Remove from package.json when convenient.
+  ckeditor: {
+    enabled: false
   },
   'split-layout-type-picker': {
     enabled: true,
--- a/cms/config/plugins.ts
+++ b/cms/config/plugins.ts
@@ -2,14 +2,11 @@ import { MAX_MEDIA_BYTES } from '../src/utils/uploadLimits'
 
 export default () => ({
   // Community plugin used by all rich-text fields (`plugin::ckeditor5.CKEditor`).
+  // Do not also enable the official `@ckeditor/strapi-plugin-ckeditor` package
+  // (plugin name `ckeditor`) — loading both double-registers core and throws
+  // `ckeditor-duplicated-modules` in the admin.
   ckeditor5: {
     enabled: true
-  },
-  // Official package ships a second CKEditor build under plugin name `ckeditor`.
-  // Leaving it enabled double-loads modules and crashes the admin with
-  // `ckeditor-duplicated-modules`. Remove from package.json when convenient.
-  ckeditor: {
-    enabled: false
   },
   'split-layout-type-picker': {
     enabled: true,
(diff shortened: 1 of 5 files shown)


Infi-Knight (commented):
Ran this locally on two worktrees (staging vs this branch, separate ports, copy of my local DB) to see the before/after.

On staging the admin is blank and the throw is:

```
CKEditorError: ckeditor-duplicated-modules
  at /admin/node_modules/.strapi/vite/deps/@ckeditor_ckeditor5-markdown-gfm.js:439:9
```

It doesn't show up in the console normally, you only catch it if you hook `window.onerror` before the bundle loads, which is probably why this was annoying to track down. Vite is pulling in both `@_sh_strapi-plugin-ckeditor_strapi-admin.js` and `@ckeditor_strapi-plugin-ckeditor_strapi-admin.js`, so the diagnosis holds up.

On this branch the admin loads clean. I opened the Grantmaking FAQ entry and poked at the live editor: `PasteFromMarkdownExperimental` is registered, the data processor is still `MarkdownGfmDataProcessor`, and `heading.options` is `paragraph, heading2..heading6`, so the no-H1 preset survived the import rewrite. Pasted `- a\n- b\n\n**bold**` and got a real bulleted list and bold in the model. Steps 2 to 4 of your manual test pass.

Three small things:

1. `config/plugins.ts:9` says "Remove from package.json when convenient", but this PR already removes it. And with the package gone, `ckeditor: { enabled: false }` isn't disabling anything, since Strapi builds its plugin list from `config.info.dependencies`. I'd just delete the block. It doesn't error on boot, I checked, so keeping it as a guard against someone re-adding the official plugin is defensible, but then the comment should say that instead.

2. `ckeditor5` isn't in `cms/package.json`. It only resolves because of `node-linker=hoisted` in `cms/.npmrc`. That was fine while it was a type-only import, but we're importing a runtime value from it now. Given the whole point here is "always import from `ckeditor5`", can we add `"ckeditor5": "47.4.0"` explicitly so it can't drift from whatever the plugin resolves and land us back here?

3. The vitest pin looks unrelated to the rest. It's also only on the cms side, root `package.json` still has `^4.0.18`, so the two importers disagree. If the vite 5.4 incompatibility is real let's pin both, otherwise `^4.0.18` plus the lockfile already satisfies `--frozen-lockfile`. Either way it probably wants its own PR.

Minor: lint and format both passed in CI, so you can tick those boxes. And could you link the Linear issue?

The comment in `app.tsx` explaining why the imports have to come from one place is good, keep that.

Infi-Knight:
Adding "ckeditor5": "47.4.0" to cms/package.json is one line, and it guards against a repeat of the exact bug being fixed here. Right now the import only resolves through hoisting, so if the @_sh plugin bumps its ckeditor5 peer, app.tsx silently follows it

Infi-Knight (approved):
left some comments for cleanup, thanks for the fix!

PR #479, opened 2026-08-07
fix(cms): keep locale and localizes on Spanish blog MDX export [INTORG-1035]

JonathanMatthey:
## Summary

Saving a Spanish foundation blog post in Strapi re-exported MDX without `locale: es` and `localizes`, which broke EN/ES pairing for sync and language switching.

The export path now stamps the requested locale on re-fetch, always writes `locale`, resolves the English pathSlug for `localizes` (with a shared pathSlug fallback when localizations lack pathSlug), and places `locale` / `localizes` after `featured` to match existing blog MDX field order.

## Related Issue

Fixes INTORG-1035

## Manual Test

1. Restart Strapi so the blog lifecycle reloads.
2. Open an existing Spanish blog post (e.g. Call for Papers ES), make a small edit, publish/save.
3. Confirm the MDX under `src/content/foundation-blog-posts/es/` still has:
   - `locale: 'es'` (or `locale: es`)
   - `localizes: '<english-path-slug>'`
   - both fields after `featured`, before `featureImage`
4. Confirm language switch / sync still matches EN and ES via `localizes`.

## Checks

- [x] Unit tests: `cms` `blogLifecycle.test.ts` (12 passing)
- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused

### DIFF
--- a/cms/src/utils/blogLifecycle.test.ts
+++ b/cms/src/utils/blogLifecycle.test.ts
@@ -153,7 +153,7 @@ describe('resolveBlogMdxFilename — write/delete path parity', () => {
 describe('generateBlogMDX — locale and localizes', () => {
   it('always writes locale for English posts', () => {
     const mdx = generateBlogMDX(makePost({ locale: 'en' }))
-    expect(mdx).toMatch(/^locale:\s*'en'/m)
+    expect(mdx).toMatch(/^locale:\s*en$/m)
     expect(mdx).not.toMatch(/^localizes:/m)
   })
 
@@ -165,8 +165,9 @@ describe('generateBlogMDX — locale and localizes', () => {
         localizations: [{ pathSlug: 'test-post', locale: 'en' }]
       })
     )
-    expect(mdx).toMatch(/^locale:\s*'es'/m)
-    expect(mdx).toMatch(/^localizes:\s*'test-post'/m)
+    // Bare YAML (no quotes) — matches checked-in blog MDX / pathSlug style.
+    expect(mdx).toMatch(/^locale:\s*es$/m)
+    expect(mdx).toMatch(/^localizes:\s*test-post$/m)
   })
 
   it('places locale and localizes after featured, before images', () => {
@@ -199,8 +200,8 @@ describe('generateBlogMDX — locale and localizes', () => {
         localizations: []
       })
     )
-    expect(mdx).toMatch(/^locale:\s*'es'/m)
-    expect(mdx).toMatch(/^localizes:\s*'test-post'/m)
+    expect(mdx).toMatch(/^locale:\s*es$/m)
+    expect(mdx).toMatch(/^localizes:\s*test-post$/m)
   })
 
   it('uses the explicit englishSlug option for localizes', () => {
@@ -212,7 +213,7 @@ describe('generateBlogMDX — locale and localizes', () => {
       }),
       { englishSlug: 'english-path' }
     )
-    expect(mdx).toMatch(/^localizes:\s*'english-path'/m)
+    expect(mdx).toMatch(/^localizes:\s*english-path$/m)
   })
 
   it('writes locale and localizes after stamping a missing document locale as es', () => {
@@ -227,8 +228,8 @@ describe('generateBlogMDX — locale and localizes', () => {
     const mdx = generateBlogMDX({ ...postFromStrapi, locale })
 
     expect(locale).toBe('es')
-    expect(mdx).toMatch(/^locale:\s*'es'/m)
-    expect(mdx).toMatch(/^localizes:\s*'test-post'/m)
+    expect(mdx).toMatch(/^locale:\s*es$/m)
+    expect(mdx).toMatch(/^localizes:\s*test-post$/m)
   })
 })
 
--- a/cms/src/utils/blogLifecycle.test.ts
+++ b/cms/src/utils/blogLifecycle.test.ts
@@ -79,12 +79,25 @@ describe('resolveBlogEnglishSlug', () => {
     ).toBe('english-slug')
   })
 
-  it('falls back to the post pathSlug when localizations are empty', () => {
+  it('returns undefined when no EN counterpart is known', () => {
+    // Must not invent localizes from the ES pathSlug alone.
     expect(
       resolveBlogEnglishSlug(
-        makePost({ locale: 'es', pathSlug: 'shared-slug', localizations: [] })
+        makePost({ locale: 'es', pathSlug: 'es-only-slug', localizations: [] })
       )
-    ).toBe('shared-slug')
+    ).toBeUndefined()
+  })
+
+  it('ignores non-en localization entries (not a real EN counterpart)', () => {
+    expect(
+      resolveBlogEnglishSlug(
+        makePost({
+          locale: 'es',
+          pathSlug: 'es-slug',
+          localizations: [{ pathSlug: 'fr-slug', locale: 'fr' }]
+        })
+      )
+    ).toBeUndefined()
   })
 })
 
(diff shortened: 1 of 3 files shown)


cms/src/utils/blogLifecycle.ts
```diff
@@ -116,8 +123,44 @@ function generateFilename({
...
+
+  const localizations = post.localizations ?? []
+  const enEntry = localizations.find(
+    (entry) => (entry.locale || '').trim() === defaultLang
+  )
+  const fromEn = enEntry?.pathSlug?.trim()
+  if (fromEn) return fromEn
+
+  const fromAny = localizations
+    .find((entry) => entry.pathSlug?.trim())
```
Infi-Knight: Minor, and only a future trap: this branch grabs the first localization with any slug at all, regardless of locale, even though the function is meant to return the English one.

Fine while we're en/es only. Add a third locale and an ES post whose EN entry is missing its `locale` field quietly picks up the FR slug, which then drives both the filename and `localizes`. The `post.pathSlug` fallback below already covers this case, so the branch could just go.

cms/src/utils/blogLifecycle.ts
```diff
@@ -144,13 +187,19 @@ export function generateBlogMDX(post: BlogResult) {
...
+  // Stable order matching existing blog MDX (locale/localizes before images).
   const frontmatterLines = [
     `title: ${yqs(post.title)}`,
     `description: ${yqs(post.description)}`,
     `date: ${post.date}`,
     post.lastUpdated ? `lastUpdated: ${post.lastUpdated}` : null,
     `pathSlug: ${post.pathSlug}`,
     `featured: ${post.featured ?? false}`,
+    // Always write locale so ES files never lose `locale: es` on re-export.
+    `locale: ${yqs(locale)}`,
```
Infi-Knight: Heads up on churn: `yqs(locale)` gives `locale: 'es'` and `localizes: 'my-post'`, but every checked-in blog file uses the bare form (`locale: es`), and Prettier's YAML formatter leaves quoting alone.

So the first re-export of each post produces two lines of diff unrelated to whatever was actually edited, and the git-sync lifecycle auto-commits it. `pathSlug` right above is deliberately unquoted. If the plan is to converge on quoted values across content types that's fine, profiles and pages already quote, just worth doing on purpose rather than as a side effect.

cms/src/utils/blogLifecycle.ts
```diff
@@ -144,13 +187,19 @@ export function generateBlogMDX(post: BlogResult) {
...
+    // Locale files must keep `localizes` pointing at the English pathSlug so
+    // sync-mdx can match translations.
+    isLocalized && englishSlug ? `localizes: ${yqs(englishSlug)}` : null,
```
Infi-Knight: Because `resolveBlogEnglishSlug` falls back to the post's own slug, an ES post with no published EN counterpart (`enPost` null, `localizations` empty) now exports `localizes: <its-own-es-slug>`. Previously it got no `localizes` line at all. I saw this in the generated output.

Two knock-ons:

- `syncUnmatchedLocales` (`cms/scripts/sync-mdx/syncOperations.ts:178`) logs `Looking for English post in Strapi with pathSlug: <es-slug>` on every sync.
- `translationMap.buildMap` (`src/utils/main/translationMap.ts:53-56`) writes `map[esSlug] = { en: esSlug, es: esSlug }`, so we advertise an EN URL with no route behind it.

Nothing visibly breaks today since ES detail routes come off the EN entries, but the field is now saying something untrue. Only emitting `localizes` when we actually found an EN counterpart would be safer.

cms/src/utils/blogLifecycle.ts
```diff
@@ -206,21 +251,24 @@ export function generateBlogMDX(post: BlogResult) {
...
+  const locale = (post.locale || defaultLang).trim() || defaultLang
+  const resolvedEnglishSlug = resolveBlogEnglishSlug(post, englishSlug)
   const filename = generateFilename({
     date: post.date,
-    pathSlug: resolveFilenameSlug(
-      post.locale,
-      post.pathSlug,
-      post.localizations?.[0]?.pathSlug
-    )
+    pathSlug: resolveFilenameSlug(locale, post.pathSlug, resolvedEnglishSlug)
```
Infi-Knight: The write path got fixed here, but its sibling in `deleteMDXFile` (line 289) didn't come along. That one still does `resolveFilenameSlug(post.locale, post.pathSlug, post.localizations?.[0]?.pathSlug)`, and `afterDelete` (~478) hands `event.result` straight to `getOutputPath(result.locale)`.

Strapi doesn't populate relations on lifecycle results, so `result.localizations` is `undefined` there, and `result.locale` is the exact field this PR argues can go missing.

So: delete an ES post whose `pathSlug` differs from the EN slug, and we unlink `es/<date>-<es-slug>.mdx` while the file actually on disk is `es/<date>-<en-slug>.mdx`. The ENOENT gets swallowed, we return `null`, the orphan stays in git, and the next `pnpm sync:mdx` pulls it back into Strapi. If `result.locale` is missing it's worse, because `getOutputPath(undefined)` aims at the EN root.

Can we run the delete through `resolveBlogEnglishSlug` and a refetched post the same way this does?

Infi-Knight (commented):
Pulled this down and ran it: `cms` tests pass 12/12, `tsc --noEmit` is clean, and I printed the actual generated frontmatter for a normal ES post and for an ES-only post.

The core fix is right. New field order matches every checked-in blog MDX I sampled, so no ordering churn, and all the `es/` files we have today already carry both fields, so nothing to backfill.

Six things inline. The two on `writeMDXFile`/`exportAllBlogLocales` I'd want sorted before merge; the rest are your call.

cms/src/utils/blogLifecycle.ts
```diff
@@ -308,8 +357,9 @@ export function createBlogLifecycle({ outputDir }: { outputDir: string }) {
       }
       try {
         const filepath = await writeMDXFile({
-          outputPath: getOutputPath(post.locale),
-          post
+          outputPath: getOutputPath(locale),
+          post: { ...post, locale },
```
Infi-Knight: `{ ...post, locale }` overwrites whatever locale the fetched document actually reports, and `getOutputPath(locale)` above does the same for the directory. `fetchBlogPost` already handles the "Strapi omitted locale" case with `post.locale?.trim() || locale`, so this second override only removes the safety net.

If `fetchBlogPost(documentId, 'es')` ever hands back the EN document (i18n fallback, i18n switched off on the content type, some future config change), the old code wrote it into the EN directory as `locale: en` and we got a harmless duplicate. Now it writes English content to `es/<date>-<slug>.mdx`, stamps `locale: 'es'`, and gives it a `localizes` pointing at itself. `paginateAllPosts` filters on `entry.data.locale === lang`, so that lands on the Spanish listing page as a real post.

I'd trust the locale we got back and log a warning when it doesn't match what we asked for. Same pattern at 392 and 457.

cms/src/utils/blogLifecycle.test.ts
```diff
@@ -22,6 +22,137 @@ function makePost(overrides: Record<string, unknown> = {}) {
...
+        locale: 'es',
+        pathSlug: 'es-only-slug',
+        localizations: []
+      }),
+      { englishSlug: 'english-path' }
+    )
+    expect(mdx).toMatch(/^localizes:\s*'english-path'/m)
+  })
+
+  it('still writes locale and localizes when locale was missing on the post', () => {
```
Infi-Knight: This one is named for the regression but asserts the old behavior: it builds a post with `locale: undefined`, then asserts `locale: 'en'` and no `localizes`. That's the bug, not the fix.

Meanwhile the actual repair lives in `fetchBlogPost`'s stamp (blogLifecycle.ts:101-108) and has no coverage at all. The second half of the test re-checks the same ES post the test above it already covers.

Worth flipping so the thing we broke is genuinely guarded. Also "omiting" on the next line.

JonathanMatthey:
all resolved - please review again

Infi-Knight (approved):

PR #483, opened 2026-08-10
feat(blocks): add CTA Buttons block [INTORG-907]

Infi-Knight:
## Summary

This PR adds a call-to-action (CTA) button block. Editors can add it to the dynamic zone of a foundation, summit, or hackathon page. The block holds one or two buttons. A pair such as "Apply now" and "Read the guide" becomes one block, not two stacked blocks.

### The block

`blocks.cta-buttons` holds a repeatable `shared.cta-button`, with a minimum of 1 and a maximum of 2. Each button has `text`, `link`, `style`, `external`, and `document`. The `style` field is primary or secondary.

The buttons are a static array prop, not child elements. Every field is plain text, so the block does not need a child component. This follows `NumberTiles`.

Three rules do not fit in the schema:

1. The block holds one or two buttons.
2. Two primary buttons are not permitted.
3. The primary button comes first.

These rules live in `cms/src/utils/ctaButtons.ts`. Both write paths use them. The MDX importer raises `INVALID_PROP_VALUE`. The admin path goes through the serializer, which raises `SerializerFieldError` with an indexed path, for example `content[i].buttons[j].style`. The admin shows the message on the field that caused the error.

One module holds these rules for a reason. Two write paths that each keep a copy of a rule drift apart.

The `external` and `document` flags are mutually exclusive. Card-grid CTAs already apply this rule in `card-grid.serializer.ts`, and the admin help text already states it. This PR adopts that rule instead of adding a second answer.

### Two supporting changes

`resolveCtaLink` in `src/utils/main/cta.ts` holds the icon rule for each kind of link. `TitleCard`, `ResourceCard`, and `NavigationCard` each kept a copy of this rule. The three call sites keep their behavior, because the `internalIcon` parameter carries the default that each one had. The site looks the same.

This change also corrects a bug. All three cards tested the URL with `isExternalHref`, which matches only `http` and `https`. A `mailto:` or `tel:` href became `/mailto:...`. The helper tests the URL with `hasUrlScheme` instead. Tests cover this.

`buttonVariants` gets a third `mode` value: `auto`. Its compound emits `dark:` versions of the dark styles. A block in a dynamic zone cannot know which site it is on, and the summit and hackathon sites are always dark. With `auto`, the secondary button follows `data-theme` and the block does not detect the site.

The `light` and `dark` values do not change. `HomepageHero` passes `mode="dark"` on a page with a light theme. The preview route shows a light button and a dark button together. A CSS-driven dark mode breaks both of these.

### The spacing

Radu specified a top margin only: 24px on mobile, 24px on tablet, 32px on desktop. Nothing at the bottom, so whatever follows supplies its own space. Those three values match spacing tokens the site already has.

The block renders on two paths and had to match on both. In an exported MDX page it sits inline, so its own margin is the only spacing it gets. In the Strapi preview it sits inside `DynamicZone`, which put a fixed 60px between every block. Both together would show 92px in the preview against 32px on the live page.

So `DynamicZone` moves its gap off the flex container and onto each item as `mt-4xl`, and skips it for blocks that own their spacing. The swap is equivalent, because `mt-4xl` on every item after the first is what `gap-4xl` already produced. No other block changes. A new `OWNS_ITS_SPACING` set names the exceptions, and this block is the first member.

Add a block to that set only if it renders on **both** paths. A preview-only block still wants the standard gap.

### The arrow icon

In this block an internal link gets an arrow. This applies to the primary button and to the secondary button. Sarah and Radu made this decision.

The decision covers custom components only. Therefore `CtaButtons.astro` sets the arrow, and the shared helper keeps `null` as its default. Template pages and the home page keep their own Figma specification.

## Related Issue

Fixes INTORG-907

This PR also delivers INTORG-877. Sarah marked that issue as a duplicate of this work. The block renders what its mock shows: two inline secondary buttons, with the icons on the right.

The sub-issue INTORG-938 is the site-wide CTA sweep. It is not in this PR. It builds on `resolveCtaLink` and comes later.

## Manual Test

1. Run `pnpm dev`. Open `/demo-foundation-page`.
2. Look near the end of the page. Three demos are there: one primary button, a primary button with a secondary button, and two secondary buttons with a download icon and an external-link icon.
3. Set the viewport width to 390px. Both button styles go to full width and stack.
4. Open `/demo-summit-page` or `/demo-hackathon-page`. The secondary button uses its dark style, because those layouts set `data-theme="dark"`.
5. Open `/preview/ui-components`. A row named "Secondary (lg, auto mode)" sits next to the explicit light row and the explicit dark row.
6. Open the Strapi admin. Add a CTA Buttons block to a foundation page. Save two primary buttons. The save fails, and the message appears on the Style field that caused the error.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

I also ran these: the site suite passes with 271 tests, the CMS suite passes with 1018 tests (61 of them are new, across four CTA files), `astro build` completes, and `sync:mdx --dry-run` reports 0 errors.

## PR Checklist

- [x] Linked issue included

This PR touches 39 files, which is more than the target. A new Strapi block needs a schema, a handler, a serializer, admin configuration, three zone schemas, an Astro component, MDX component maps, demo content, and generated types. Each of those parts needs tests. The six commits are separate, so you can review them one at a time.

## Review

Jonathan reviewed this and found one bug and three smaller points. All four were correct and all four are fixed in `fd4cf636`. Each thread has a reply.

The bug is worth repeating here, because it is not only this branch. Every card renders its CTA twice: a visible button, and an invisible overlay anchor that makes the whole card clickable. The overlay received the target attributes but never a `download` attribute, so on a document CTA the button downloaded the file and the card body navigated to it. **Staging has the same bug today**, on every document card. The filename derivation was private to `LinkButton`, which is why the overlay could not have it. It now lives in `resolveCtaLink`, and both the button and the overlay use it.

The other three: the serializer validated composition before fields, so an empty draft was told "Two buttons cannot both be primary" when the real problem was that neither button had text. The MDX handler stored untrimmed values while validating trimmed ones. And `resolveCtaLink` typed `icon` as nullable, which made `NavigationCard` inconsistent with its two siblings; an overload now types it as non-null whenever the caller names an internal icon.

Tests: 5 for the download attributes, 2 for the validation order, 1 for the trim. Each one fails against the code as it was.


src/components/shared/NavigationCard.astro
```diff
@@ -24,28 +24,27 @@ const {
...
-const href = isAbsoluteHttp ? buttonUrl : ensureLeadingSlash(buttonUrl)
 
-const overlayTargetAttrs = external
-  ? ({ target: '_blank', rel: 'noopener noreferrer' } as const)
-  : {}
+// Figma default uses the bidirectional swap icon; external/document override.
+const {
+  href,
+  external,
+  icon: trailingIcon,
```
JonathanMatthey, line 32: `trailingIcon` is typed as `CtaIconName | null`, but NavigationCard always renders `<Icon name={trailingIcon} … />` with no null guard. 

Either keep the unconditional render and narrow the type

cms/src/serializers/blocks/cta-buttons.serializer.ts
```diff
@@ -0,0 +1,78 @@
...
+      path: ['buttons', index, 'document']
+    })
+
+  return fieldErrors
+}
+
+export function serialize(block: { buttons?: CtaButtonEntry[] }): string {
+  // Strapi's required/min/max constraints aren't enforced on save, and the
+  // composition rules can't be expressed in the schema at all.
+  const ruleError = validateCtaButtonComposition(block.buttons)
```
JonathanMatthey: Composition validation runs before per-button field validation. An incomplete draft with two empty buttons (both defaulting to primary via `style ?? 'primary'`) fails first with “Two buttons cannot both be primary” on `buttons[1].style`

Run `validateButton` (field errors) first and only run `validateCtaButtonComposition` when every button has enough shape to judge style, or skip incomplete buttons when counting primaries. Alternatively, collect both classes of errors and return them together.

cms/scripts/sync-mdx/ctaButtonsHandler.ts
```diff
@@ -0,0 +1,117 @@
...
+    const raw = getStaticLiteralAttr(node, 'buttons', { required: true })
+
+    if (!Array.isArray(raw) || !raw.every(isRawButton)) {
+      fail(
+        'Prop "buttons" must be an array of { text, link, style?, external?, document? } objects. ' +
+          '"text" and "link" are required and non-empty; "style" is "primary" or "secondary".'
+      )
+    }
+
+    const buttons: CtaButton[] = (raw as RawButton[]).map((entry) => ({
```
JonathanMatthey, line 91: The parser accepts non-empty `text`/`link` via `trim()` for validation but stores the untrimmed values. 

Mirror NumberTiles: `text: entry.text.trim()`, `link: entry.link.trim()` when mapping into `CtaButton[]`.

src/components/shared/TitleCard.astro
```diff
@@ -23,15 +23,18 @@ const {
...
-// Absolute http(s) URLs must not get a leading slash (e.g. partner PDFs).
-const isAbsoluteHttp = isExternalHref(buttonUrl)
-const external = buttonExternal || isAbsoluteHttp
-const href = isAbsoluteHttp ? buttonUrl : ensureLeadingSlash(buttonUrl)
+const {
+  href,
+  external,
+  icon: trailingIcon,
+  // Mouse-only "whole card is clickable" overlay shares the button's target.
+  targetAttrs: overlayTargetAttrs
```
JonathanMatthey: The whole-card overlay still only receives `targetAttrs` from `resolveCtaLink`, not a `download` attribute. Clicking the card body on a document CTA navigates to the file instead of downloading, while the visible LinkButton sets `download` via `document={buttonDocument}`. 

extend `resolveCtaLink` to expose download attrs and spread them onto the overlay `<a>`

JonathanMatthey (commented):
1 bug, 3 small things

src/components/shared/TitleCard.astro
Infi-Knight: Good catch, and this one predates the branch. Staging has the same half-configured overlay, so the bug is live today on every document card, not only here. Extracting `resolveCtaLink` was the moment to catch it and I did not.

Fixed as you suggested. The filename derivation moved out of `LinkButton` into `resolveCtaLink`, which now returns `downloadAttrs` next to `targetAttrs`. All three cards spread both onto the overlay, and `LinkButton` uses the same helper, so there is one derivation instead of two.

Verified in the browser: the button and the overlay now carry the same `download="2025-application-guide.pdf"`. Added 5 tests for `downloadAttrs`, including the query-string and no-filename cases. In fd4cf636.

cms/src/serializers/blocks/cta-buttons.serializer.ts
Infi-Knight: Agreed, and the failure mode is worse than a confusing message: the editor is pointed at a Style dropdown they never touched.

Fixed by splitting the count check from the style check. `validateCtaButtonCount` runs first, because the serializer needs a usable array before it inspects entries. Then the per-field errors. Then `validateCtaButtonStyles`. `validateCtaButtonComposition` still exists as both halves in order, since the MDX parser has already rejected empty entries by the time it calls that.

Two tests: two empty buttons now report the four missing text and link paths, and a genuine two-primary clash still reports `buttons[1].style`. In fd4cf636.

cms/scripts/sync-mdx/ctaButtonsHandler.ts
Infi-Knight: Correct, and `numberTilesHandler` was open in front of me as the reference pattern. Fixed: `text` and `link" are trimmed when mapping into `CtaButton[]`.

Worth noting why it matters beyond tidiness. The serializer trims on the way out, so untrimmed values in Strapi would come back trimmed on the next export and show up as a spurious diff in the MDX. Test added. In fd4cf636.

src/components/shared/NavigationCard.astro
Infi-Knight: Right, and it is my regression. Before the refactor this card computed `trailingIcon` inline, so it was always a string and the unconditional render was safe. Widening the return type to `CtaIconName | null` is what made the call site inconsistent with its two siblings.

Took your first option and narrowed the type. `resolveCtaLink` now has an overload: when the caller passes `internalIcon`, every link kind resolves to an icon, so `icon` is `CtaIconName` and never null. The unconditional render stays, and the false null disappears at both call sites that name an internal icon, here and in `CtaButtons`.

`astro build` is the typecheck and it passes. In fd4cf636.

JonathanMatthey (approved):
approved,

some tiny things like - but leave them... 
- [src/utils/main/cta.ts:66](vscode-webview://0ft6h3nm321cpi0em5d1hgtg3v5gapqf93mfjv7bimljqtlb3kjd/src/utils/main/cta.ts#L66) — ?? href is unreachable dead code
- [src/components/blocks/CtaButtons.astro:41-49](vscode-webview://0ft6h3nm321cpi0em5d1hgtg3v5gapqf93mfjv7bimljqtlb3kjd/src/components/blocks/CtaButtons.astro#L41-L49) — double .trim() on text and link

JonathanMatthey:
look great

dunno if we want bottom margin, would have to check figma... 🤷 

image


image

Infi-Knight:
> dunno if we want bottom margin, would have to check figma... 🤷

Radu settled this in Slack a couple of hours before you looked. A CTA belongs to the paragraph above it, so 24px above and 32px from desktop, and the block below gets the normal 120px.

PR #484, opened 2026-08-10
fix(cta-strip): restore the secondary CTA button [INTORG-908]

Infi-Knight:
## Summary

PR #458 made the CTA strip purple only. It also removed the secondary button. Only the color change was wanted. Sarah confirmed on 2026-08-10 that the secondary button is still necessary. This PR brings it back. Strips stay purple only, and the `color` field does not return.

The restored markup is the same as the version before #458. I compared it against `daeff223^`. The secondary button carries no icon, which is what it had before. That is correct, because the CTA strip is a template surface. Sarah scoped the new arrow-on-internal-link rule to custom components only.

### The two fields travel together

A label with no href is a dead button. An href with no label is unreadable. So the handler, the serializer, and the Astro component all need both fields. The validator now tells the editor which half is missing. Before, the button disappeared without a message.

Absent secondary fields go to Strapi as `null`, not as omitted keys. This matches `heading` and `description`. A PUT then clears a secondary button that an editor removed from MDX. Without this, the last value stays in place.

### A contrast bug came with it

The browser check found the restored button rendering white on light indigo, at about 1.06:1, on the summit and hackathon pages.

The microsite content wrapper styles every descendant anchor with `[&_a]:text-current`. That compiles to a descendant selector, which outranks the button's own `text-neutral-100`. So the white body color of the page reached inside the strip card. The card is always light, whatever the page theme is.

This bug is not new. `daeff223^` shows that the summit demo page had the same secondary button before #458, and that `purple` already meant `bg-soft-indigo-50`. PR #458 hid the bug when it deleted the button. This PR makes it visible again, so this PR fixes it.

The fix stays inside `CtaStrip.astro`. The card re-applies the light-mode colors through a selector that outranks the wrapper. Nothing outside that one component can change.

**One related problem is left alone.** The same wrapper also carries `[&_a]:underline`, so it underlines buttons. That affects the primary button on staging today, so it is not this PR's to change. It belongs to the INTORG-938 sweep. The general shape is worth stating: a prose wrapper that styles every descendant anchor also styles every button. `LinkButton` emits `data-component="LinkButton"`, which is the hook for excluding buttons at the source.

### The generated types

I applied the type change by hand. `strapi ts:generate-types` boots Strapi and reads the component schemas from the local database. That database still holds the CTA Buttons component from the INTORG-907 branch, so a full run emits `BlocksCtaButtons` as well. Check `git diff` on generated files when you regenerate on a branch.

## Related Issue

Fixes INTORG-908

Sarah asked Jonathan to do this work in Slack. I did it instead, because it sits next to the INTORG-907 CTA button work I already had open. Jonathan, please do not duplicate it.

This PR is independent of #483 (INTORG-907). Either one can merge first. Three files overlap, and the changes sit in different parts of each file.

## Manual Test

1. Run `pnpm dev`. Open `/demo-foundation-page`. The CTA strip shows a purple primary button and an outlined secondary button.
2. Open `/summit/demo-summit-page` and `/hackathon/demo-hackathon-page`. Both strips show the secondary button. Its text is dark, and it is readable on the light card.
3. Open `/grant/demo-grant-page`. The "View all grants" secondary button appears. On staging it does not, because the handler drops it.
4. Remove `secondaryButtonLink` from one demo file and keep `secondaryButtonText`. The button does not render, and the sync reports which field is missing.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

I also ran these: the site suite passes with 247 tests, the CMS suite passes with 959 tests, `astro build` completes, and the MDX formatter reports no changes.

## PR Checklist

- [x] Linked issue included
- [x] Scope is focused

## Note for reviewers

`demo-grant-page.mdx` and `demo-grant-overview-page.mdx` already carry `secondaryButtonText`, `secondaryButtonLink`, and `color` on staging. PR #458 took those fields out of the schema, but it left the demo content alone. The handler drops all three today without a message.

Two results follow. First, this branch makes the two secondary fields work again, so the sync now imports content that staging discards. Second, `color` stays ignored, which is correct. Removing the dead `color` attribute from those two files belongs to whoever next edits them.

The demo strips on the foundation, summit, and hackathon pages had no secondary button, so this PR adds one to each. Those three pages use the MDX path, which the grant demos do not exercise.


cms/src/utils/contentValidation.ts
```diff
@@ -371,6 +373,27 @@ export function validateCtaStrip(
       path: ['ctaStrip', 'primaryButtonLink']
     })
   }
+
+  // The secondary CTA is optional, but half of one is not: a label with no
+  // href is dead and an href with no label is unreadable, so flag the gap
+  // rather than silently dropping the button at render time.
+  const isFilled = (value: unknown) =>
+    typeof value === 'string' && value.trim() !== ''
+  if (isFilled(secondaryButtonText) && !isFilled(secondaryButtonLink)) {
```
JonathanMatthey, line 382: Add cases for (a) text-only secondary → error on `secondaryButtonLink`, (b) link-only → error on `secondaryButtonText`, (c) both set → success, (d) whitespace-only halves treated as empty.

src/schemas/content.ts
```diff
@@ -120,7 +120,9 @@ const grantCtaStripSchema = z.object({
   heading: z.string().optional(),
   description: z.string().optional(),
   buttonText: z.string(),
-  buttonLink: z.string()
+  buttonLink: z.string(),
+  secondaryButtonText: z.string().optional(),
+  secondaryButtonLink: z.string().optional()
```
JonathanMatthey: Add a `.superRefine` (or equivalent) so both secondary fields are required together when either is non-empty, matching the Strapi validator and lifecycle export rule.

cms/scripts/sync-mdx/ctaStripHandler.ts
```diff
@@ -48,6 +52,13 @@ async function handleCtaStrip(
     if (heading) block.heading = heading
     if (description) block.description = description
 
+    // A half-specified secondary button would render as a dead or unlabelled
+    // control, so it only survives when both halves are present.
+    if (secondaryButtonText && secondaryButtonLink) {
```
JonathanMatthey: Align the handler with serializer/renderer: require both `secondaryButtonText?.trim()` and `secondaryButtonLink?.trim()` before assigning onto the block.

cms/scripts/sync-mdx/ctaStripHandler.test.ts
```diff
@@ -31,11 +31,37 @@ describe('CtaStrip handler', () => {
...
         primaryButtonText: 'Stay in touch',
-        primaryButtonLink: '/contact'
+        primaryButtonLink: '/contact',
+        secondaryButtonText: 'Get involved',
+        secondaryButtonLink: '/get-involved'
       }
     ])
   })
 
+  it('drops a half-specified secondary CTA', async () => {
```
JonathanMatthey: Keep one comprehensive test (or fold link-only into the existing case) to reduce noise.

src/schemas/content.ts
JonathanMatthey: cross-field rule → .superRefine

JonathanMatthey (approved):
no blockers, just improvements, will let you decide which to implement and you can merge!

nice job

src/schemas/content.ts
Infi-Knight: Added, as a `superRefine` on `grantCtaStripSchema`. It flags the missing half and anchors the message on that field, so a text-only pair points at `secondaryButtonLink` and a link-only pair points at `secondaryButtonText`. Whitespace counts as empty, matching the other four places that judge this pair.

6 tests in `src/schemas/content.test.ts`: both halves, neither half, each half alone, one whitespace half, and two whitespace halves. In f89af1c7.

cms/scripts/sync-mdx/ctaStripHandler.ts
Infi-Knight: Fixed. The handler now tests the trimmed values and stores them trimmed, so `secondaryButtonText="   "` no longer reaches Strapi only to be dropped again on the way out.

Same problem existed in three more places on this branch, all now fixed: both grant lifecycles compared untrimmed values when deciding whether to export the secondary, and the podcast export did the same. In f89af1c7.

cms/scripts/sync-mdx/ctaStripHandler.test.ts
Infi-Knight: Folded into one `it.each` covering text-only, link-only, whitespace text, whitespace link, and both whitespace. Every case now asserts that **neither** field survives, rather than checking one field per case as it did before.

Added a separate test that the pair it does keep is stored trimmed. In f89af1c7.

cms/src/utils/contentValidation.ts
Infi-Knight: Added all four, plus two more. `validateCtaStrip` had the pair rule but no test for it at all.

Cases: both halves set, neither set, text-only flagged on `secondaryButtonLink`, link-only flagged on `secondaryButtonText`, one whitespace half flagged, and two whitespace halves treated as no secondary at all. In f89af1c7.

src/components/blocks/CtaStrip.astro
```diff
@@ -58,7 +70,16 @@ if (!hasPrimary) {
...
+        'bg-soft-indigo-50',
+        // The summit and hackathon content wrapper styles every descendant
+        // anchor with `[&_a]:text-current`, so the page's white body colour
+        // reaches inside this card. That selector has a higher specificity
+        // than the button's own `text-neutral-100`, so the secondary button
+        // renders white on light indigo (about 1.06:1). The card is always
+        // light, whatever the page theme, so re-assert the light-mode colours
+        // through a selector that outranks the wrapper.
+        '[&_a.cta-strip-secondary]:text-neutral-100',
+        '[&_a.cta-strip-secondary:hover]:text-neutral-900'
```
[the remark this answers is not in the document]
Infi-Knight: Fixed, though the diagnosis was not quite right. I measured it before changing anything.

Under keyboard focus the text was `rgb(41,41,41)`, not white. The override here is `.x a.cta-strip-secondary` at specificity (0,2,1), which beats the wrapper's `.x a` at (0,1,1), so the button never falls back to `currentColor`. Contrast was about 10:1 and the 2px black outline and neutral-25 fill were both applied.

But the same specificity also beats `focus-visible:text-neutral-900` at (0,2,0), so the text stayed neutral-100 instead of deepening to black. Real, smaller than described. Added `[&_a.cta-strip-secondary:focus-visible]:text-neutral-900` and re-measured: `rgb(0,0,0)` on neutral-25 with the 2px black outline. In f89af1c7.

src/schemas/content.ts
```diff
@@ -120,7 +120,9 @@ const grantCtaStripSchema = z.object({
...
 })
```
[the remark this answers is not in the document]
Infi-Knight: Agreed, and added. See the reply on Jonathan's thread for the same point. The `superRefine` anchors the error on whichever half is missing, and treats whitespace as empty. In f89af1c7.

cms/src/api/grant-page/content-types/grant-page/lifecycles.ts
```diff
@@ -154,6 +156,13 @@ export function generateGrantPageMDX(
...
             ...(ctaStrip.description
               ? { description: ctaStrip.description }
+              : {}),
+            // Both halves or neither, matching the serializer and renderer.
+            ...(ctaStrip.secondaryButtonText && ctaStrip.secondaryButtonLink
+              ? {
+                  secondaryButtonText: ctaStrip.secondaryButtonText,
+                  secondaryButtonLink: ctaStrip.secondaryButtonLink
+                }
               : {})
```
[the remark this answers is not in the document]
Infi-Knight: Correct, fixed here and in the grant-overview lifecycle. Both now compare and write the trimmed values.

Same class of bug turned up in two more places while I was checking: the MDX handler stored untrimmed values, and the podcast export did too. All four are consistent now. In f89af1c7.

cms/src/api/grant-overview-page/content-types/grant-overview-page/lifecycles.ts
```diff
@@ -68,6 +70,13 @@ export function generateGrantOverviewPageMDX(
...
             ...(ctaStrip.description
               ? { description: ctaStrip.description }
+              : {}),
+            // Both halves or neither, matching the serializer and renderer.
+            ...(ctaStrip.secondaryButtonText && ctaStrip.secondaryButtonLink
+              ? {
+                  secondaryButtonText: ctaStrip.secondaryButtonText,
+                  secondaryButtonLink: ctaStrip.secondaryButtonLink
+                }
               : {})
```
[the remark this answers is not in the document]
Infi-Knight: Fixed, same change as the grant-page lifecycle. In f89af1c7.

Infi-Knight:
Ran this end to end against a local Strapi, MDX into Strapi and back out again, and it turned up a live bug that has nothing to do with the secondary button.

**The podcast page has not been able to sync since #458 merged.** That PR took `color` off `blocks.cta-strip`, but three places kept sending it: the podcast frontmatter schema defaulted it to `purple`, the sync transformer put it in the payload, and the MDX export wrote it back out. Strapi rejected every attempt with `Invalid key color at ctaStrip`. A real sync now reports 0 errors and creates the podcast page for the first time.

**The dry-run cannot catch this**, which is why nobody noticed. It never PUTs, so Strapi's own validation never runs, and `build-and-dryrun` in CI only runs the dry-run. Worth knowing when reading a green check on a schema change.

Four dead `color` keys came out of content as well, and one line of demo copy that promised a green variant that no longer exists.

What the run covered:

1. Strapi boots on the new schema.
2. Real sync of all 200 pages: 0 errors.
3. Seven CTA strips persisted with their secondary fields, including a `mailto:` link.
4. A save through the admin API triggered the export lifecycle, and the secondary came back out into MDX correctly.
5. The import wrote no MDX of its own, so the `x-skip-mdx-export` header is doing its job.

The demo page I used for step 4 is restored and Strapi is back in sync with it.

Infi-Knight:
Merged `staging` in to clear the conflict. One resolution needs calling out, because it partly reverses #481.

Sarah's #481 fixed the podcast CTA strip by removing `color` **and** the two secondary fields. Removing `color` was right and I kept it. Removing the secondary was also right at the time, because `blocks.cta-strip` had no secondary after #458.

This PR puts the secondary back on that component, and the podcast page uses `blocks.cta-strip`. So I restored `secondaryButtonText` and `secondaryButtonLink` on the podcast path: the frontmatter schema, the sync transformer, and the MDX export. Taking #481 wholesale would have meant an editor could set a secondary on a podcast page in the admin and have it silently dropped on export.

The restored path follows the same rule as everywhere else in this PR: both halves or neither, trimmed, whitespace counts as empty.

`color` stays gone. There is a test asserting it, so nothing puts it back.

Conflicts were in `src/schemas/content.ts`, `cms/scripts/sync-mdx/mdxTransformer.ts`, `cms/src/utils/podcastPageMdx.ts` and its test. After the merge: site 253 passing, CMS 991 passing, `astro build` completes.

PR #495, opened 2026-08-12
add a git pull before sync images

JonathanMatthey:
this adds a gitpull workflow before syncimages

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -3,6 +3,8 @@ name: Build and Sync
 # Conditionally rebuilds the Strapi CMS on the self-hosted runner if cms/ changed,
 # and/or syncs to Strapi when src/content changes (.md/.mdx), navigation JSON changes,
 # or static media under public/img/, public/uploads/, or public/documents/ changes.
+# When only media/content changes (no cms/ changes), pull-server-files runs a lightweight
+# git pull on the self-hosted runner so new files are on disk before sync-to-strapi seeds them.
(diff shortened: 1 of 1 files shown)


.github/workflows/merge.yml
```diff
@@ -199,6 +201,49 @@ jobs:
...
             core.setOutput('navigation_changed', String(navigationChanged))
 
+  # Runs a git pull on the self-hosted Strapi instance when media or content files changed but
+  # the CMS code itself did not. This makes new files (e.g. images under public/uploads/) available
+  # on disk so that sync-to-strapi's POST /api/seed-media call can register them in Strapi.
+  # Skipped when cms_changed is true because rebuild-strapi already does a git pull in that case.
+  pull-server-files:
+    needs: detect-changes
+    if: |
+      needs.detect-changes.outputs.content_changed == 'true' &&
```
Anca2022: `content_changed` also covers markdown-only changes, but `sync:mdx` never touches the VM's disk, only `seed-media` does. 

one suggestion on the trigger: we could add a narrower `media_changed` and gate this job on that instead?

Anca2022 (approved):
Nice fix for the seeding gap! :D 

Left a comment, not a blocker though, just a suggestion. 

lgtm ✅

.github/workflows/merge.yml
Anca2022: Not a bug, it just causes an unnecessary self-hosted runner dispatch

PR #493, opened 2026-08-12
ci(testing): vitest on PR and coverage thresholds

bosbaber:
## Summary
- Added automated unit testing and test coverage checking as part of the PR checks.
- Added default thresholds a couple of points lower than the current measured values. Should increase over time.
See https://linear.app/interledger/issue/INTORG-1057/ci-vitest-should-run-during-each-build-as-well-as-coverage-check

## Manual Test
See pipeline. Consider adding code to see if it breaks the pipeline.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] PR title follows Conventional Commits (e.g. `feat: ...`, `fix: ...`)
- [x] Linked issue included
- [x] Scope is focused (target ~10-20 files when possible)

## Auto generated description
This pull request introduces code coverage reporting for both the main source code and the CMS, and updates the test workflow to run these test suites independently and in parallel. It also upgrades the `vitest` testing framework and related dependencies in the CMS to the latest version, and adds the necessary coverage plugins and dependencies. Additionally, coverage output is now ignored by Prettier formatting.

The most important changes are:

**CI Workflow Improvements**
* The `test` job in `.github/workflows/pr-checks.yml` is split into two separate jobs: `test-src` (for main source code) and `test-cms` (for the CMS), so failures in one do not hide failures in the other. Both jobs now run in parallel and independently report results.
* Both test jobs now generate code coverage reports using `pnpm test:coverage`, and upload the results as build artifacts for further inspection.

**CMS Test and Coverage Enhancements**
* Adds a `test:coverage` script to `cms/package.json` to run `vitest` with coverage enabled.
[...]

PR #492, opened 2026-08-12
feat(hackathon): FoundationFooter uses hackathon navigation [INTORG-1046]

JonathanMatthey:
## Summary

- Wire `HackathonPageLayout` to `FoundationFooter` with `site="hackathon"` so the hackathon chrome uses the same footer component as foundation, driven by the hackathon nav config.
- Teach `FoundationFooter` an optional `site` prop (default `foundation`) that selects the navigation source via `getNavigation`.
- Populate EN/ES `hackathon-navigation` with Overview, FAQ, and a Resources group (participation requirements, code of conduct).

## Test plan

1. Open a hackathon page (e.g. `/hackathon/overview`).
2. Confirm the footer shows Hackathon (Overview, FAQ) and Resources (Participation requirements, Code of conduct) columns, plus the usual legal Resources column if still appended.
3. Switch locale to ES and confirm Spanish labels and `/es/hackathon/...` hrefs.
4. Confirm foundation pages still use the foundation menu (default `site`).

## Related

Fixes INTORG-1046

PR #488, opened 2026-08-11
feat(strapi): rel links helper text and auto-correct

adrianboros:
## Summary
- standardize rel links helper text messages across Strapi
- added middleware to auto-correct rel links

## Related Issue
Fixes INTORG-1038

## Manual Test
- in the Strapi verify URLs fields have a helper text present
- in the Strapi, save a Grant Overview Page with pathSlug typed as `/our-grantmaking` (with slash in front) and a CTA link typed as `grants/apply` (no slash), then reload the entry to confirm they come back as our-grantmaking and /grants/apply respectively.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] PR title follows Conventional Commits (e.g. `feat: ...`, `fix: ...`)
- [x] Linked issue included
- [x] Scope is focused (target ~10-20 files when possible)
- [ ] Screenshots for UI changes (if applicable)

PR #501, opened 2026-08-13
fix(cta): make the CTA icon follow the link kind everywhere [INTORG-938]

Infi-Knight:
## Summary

A CTA icon should follow the link kind: document downloads, external opens a
new tab, anything else is internal. Three card components did this. Six other
places did not.

`CtaLink` and `SplitLayout` showed `arrow-right` **when the link was
external**, which is the rule inverted. `PageHero`, `GrantPage` and `CtaStrip`
always showed `arrow-right`. All six now call `resolveCtaLink`.

`shared.cta-link` and `shared.primary-cta-link` gain a `document` flag, which
`shared.secondary-cta-link` already had. External and document are mutually
exclusive, checked in the MDX parser and the admin validator through the same
helper.

`Link.astro` drops its own hardcoded external SVG and uses `Icon`.

| Surface | Internal | External | Document |
| --- | --- | --- | --- |
| `CtaLink` | **gains the arrow** | `external-link` | `download` |
| `SplitLayout` | unchanged, none | `external-link` | `download` |
| `CtaStrip` | unchanged, arrow | `external-link` | not expressible |
| `PageHero` | unchanged, arrow | `external-link` | `download` |
| `GrantPage` | unchanged, arrow | `external-link` | `download` |
| `Link` | unchanged, none | `external-link` | `download` |
| `GrantFaqSection` | unchanged, none | `external-link` | `download` |

Only `CtaLink` changes its internal case. Sarah scoped arrow-on-internal to
components an editor picks. The rest are page templates.

## The PDF trap

Sarah asked how an editor links a PDF. The media library's Copy Link button
gives an absolute URL on the CMS origin, which is firewalled. Pasting it
produced a link that opened a new tab against a host that will not answer. The
auto-correct from #488 returns early on absolute URLs, so nothing caught it.

`stripUploadOrigin` reduces such a URL to its path. It matches the full
`/uploads/img/original/` prefix, so a real external link with an uploads path
is left alone. The `document` helper text now spells out the steps.

## Related Issue

Fixes INTORG-938

## Manual Test

1. `/grant/innovation/grant-web/`: the "Learn about Web Monetization" button
   shows the external-link glyph, not an arrow.
2. `/grant/education/on-campus/`: "Get Notified" now has an arrow. It had no
   icon before.
3. In the admin, paste a Copy Link URL into a CTA and save. The stored value is
   `/uploads/img/original/<file>.pdf`.
4. Tick External Link and Document Download together. The save is refused.
5. On a CTA strip, set Primary Document Download. The button downloads and
   shows the download icon. Same for the FAQ section button.

## Checks

- [x] `pnpm run format`
- [x] `pnpm run lint`

## PR Checklist

- [x] Linked issue included

## Verification

Scanned every CTA anchor on 1329 built pages, before and after, reading the
glyph from each rendered SVG path. **External links showing `arrow-right`: 16
to 0.** Four links with an absolute URL and no external flag are now correctly
external, because `resolveCtaLink` reads the href.

Two shipping schemas changed, so a dry-run is not enough. A real `--force` sync
against a local Strapi gave 0 errors, 202 updated, no MDX churn.

CMS tests 1152 pass. Includes a round-trip test for `shared.cta-link`, which
had none.

## A trap if you test this in dev

Adding a field to a content collection schema needs `.astro` and
`node_modules/.astro` deleted. Astro caches parsed frontmatter, a dev-server
restart does not clear it, and the new field arrives as `undefined` while the
button silently renders as internal.

## For the reviewer

`Link.astro` is icons only. Its pill shape, fluid padding and white-on-purple
colours stay. That hero is a different design from `PageHero`, and swapping in
`LinkButton` would put a purple button on a purple gradient. What a primary CTA
looks like on a dark hero is a question for Radu.

No page currently renders a `Link.astro` hero CTA, so that change is untested
in a browser.

`blocks.card` and `blocks.card-link` are deliberately left out. They are only
reachable through `cards-grid` and `card-links-grid`, and neither is in any
content type's dynamic zone, so no editor can add one. Anca confirmed they are
dead components.

One commit here is unrelated to the CTA work. #492 added a `Resources` group
to the hackathon navigation and left `navigation.test.ts` expecting
`Hackathon` alone. Nothing ran that test until #493 added the `Test src` job,
so `staging` has been red since, and every open PR inherited it. Fixed in
`2f3865f0` because it blocked this PR. Drop that commit if you would rather it
went in on its own.


Anca2022 (commented):
From my understanding, the task asks for **all** buttons on the website to have the same shape

for example: Cta Strip should support `download` and `external` for buttons

Image

Call to action should support `download` 

Image

All secondary internal buttons should have right arrow, the way primary buttons do


I'm also getting this error on my local, not sure where this is coming from 

Image

I asked for clarification on the task, until we're getting a response I'll pause the review :)

cms/src/index.ts
```diff
@@ -1044,12 +1044,24 @@ async function configureFieldLabels(strapi: StrapiInstance) {
...
+      document: 'Document Download'
     },
     'shared.primary-cta-link': {
       link: 'Link',
       text: 'Button Text',
-      external: 'External Link'
+      external: 'External Link',
+      document: 'Document Download'
+    },
+    'shared.cta-button': {
```
Anca2022, line 1056: this is duplicated, please remove

Image

cms/src/serializers/blocks/cta-link.serializer.ts
```diff
@@ -26,9 +27,18 @@ export function serialize(block: {
...
+  if (block.external && block.document) {
+    throw new Error(
+      'CtaLink cannot be both external and document. Pick one: external ' +
+        'opens a new tab, document downloads a file.'
+    )
+  }
   if (block.external) {
     attrs.push('external={true}')
   }
+  if (block.document) {
```
Anca2022, line 39: When adding a `Call to action button` on `Foundation Page`, the `external` flag is correctly picked up, but `download` isn't being picked up
Image

cms/src/utils/contentValidation.test.ts
```diff
@@ -396,6 +439,29 @@ describe('validateGrantPagePrimaryCta', () => {
...
+    const err = validateGrantPagePrimaryCta({
+      primaryCta: {
+        text: 'Report',
+        link: 'https://example.com/r.pdf',
+        external: true,
+        document: true
+      }
+    })
+    expect(err?.message).toBe(
+      'Primary Call to Action: Pick either External Link or Document Download, not both'
```
Anca2022, line 460: On Foundation page, Card grid, CTA

Image

Anca2022: Same for Event card - apply - CTA - the validator let's you save, but the serializer only writes `button external` to mdx

Image

Anca2022: Actually, the serializer does not write `external` at all

Anca2022 (approved):
big PR :D 

found 2 functional issues: 
- 2 components that render on the Custom pages don't serialize the `document` flag
- validation allows adding both flags (external & document) to 2 cta's

I did not review styles, as CTA consistency across the website is still in discussion and will be broken up to separate issues

Overal, great job, Ravi!

cms/src/index.ts
Infi-Knight: Removed. Both `shared.cta-button` and `blocks.cta-buttons` appeared twice in `componentLabels`. The later pair won, so nothing was broken, but it was noise.

cms/src/serializers/blocks/cta-link.serializer.ts
Infi-Knight: Good catch, and the cause was not in this file.

A hero CTA goes to MDX frontmatter, not through the `CtaLink` serializer. `heroFrontmatter` in `cms/src/utils/mdx.ts` wrote `external` and never `document`, so the checkbox had nowhere to land. `buildHeroPayload` in `mdxTransformer.ts` dropped it on the way back in too, so an editor could not have recovered the value from MDX either.

Both directions now carry it, and `HeroCta` gained the field.

cms/src/utils/contentValidation.test.ts
Infi-Knight: Both are fixed, and you are right that the validator was the hole.

`hasConflictingCtaFlags` already existed, and the CTA link, CTA buttons and card grid all called it. The hero, the event card and the split layout never did. So the helper text promised a rule that three surfaces did not enforce. All three now reject the pair, on save and on sync.

On the event card Apply CTA, `document` was missing from five places: the block type, the MDX handler, the serializer, `EventCard.astro` and `EventApply.astro`. `EventApply` also hard-coded its icon as `external-link` or `arrow-right` with no download case, so it now goes through `resolveCtaLink` like every other CTA. Its internal case keeps the arrow, since the event card is a page template.

`external` was written for the event card, at `event-card.serializer.ts:152`. I think the screenshot in your last note is the hero case, where neither flag was written. If you saw an event card drop `external` as well, tell me and I will look again.

